PR #99, opened 2026-04-17
Add reservation_requests scaffold (#6)

Psheikomaniac:
## Summary
- New migration `2026_04_17_125725_create_reservation_requests_table.php`: id, `restaurant_id` FK (cascade on delete), `source`, `status`, `guest_name`, nullable `guest_email`/`guest_phone`, `party_size` (unsignedInt), nullable `desired_at` (datetime, UTC), nullable `message`, nullable `raw_payload`, `needs_manual_review` (bool, default false), timestamps.
- `App\Models\ReservationRequest` with method-based `casts()` for `source` → `ReservationSource`, `status` → `ReservationStatus`, `desired_at` → `datetime`, `raw_payload` → `encrypted:array`, `needs_manual_review` → `boolean`, `party_size` → `integer`.
- Relations: `restaurant()` (belongsTo), `replies()` (hasMany ReservationReply), `latestReply()` (hasOne ReservationReply → `latestOfMany`).
- `ReservationRequestFactory` with default state (`status=new`, `source=web_form`), plus `inReview()`, `confirmed()` and `forRestaurant($restaurant)` helpers.
- 12 feature tests covering factory states, enum cast roundtrip, `desired_at` Carbon cast incl. null, raw-payload encryption-at-rest (`Crypt::decryptString`), DB default for `needs_manual_review`, cascade delete from restaurant, `restaurant_id` NOT NULL enforcement, and `belongsTo` navigation.

## Why
Foundation for every reservation source. A uniform record shape lets the dashboard (PRD-004), email ingestion (PRD-003), and AI reply flow (PRD-005) all read the same table.

## Deviations / notes
- **`raw_payload` column is `text`, not `json` as the PRD table literally states.** `encrypted:array` writes ciphertext, which is not a valid JSON value — a `json` column would reject it in MySQL/Postgres. Encryption at rest is the stronger requirement here (raw email bodies + form payloads contain guest PII), so the column type was downgraded to `text` to make that cast work. Functionality from the cast side is unchanged: reads/writes go through `array` in PHP.
- No `$hidden` fields on this model. Guest PII (name/email/phone) is protected at the serialization boundary via API Resources (per `CLAUDE.md` — "Validierung / Serialisierung gehören in FormRequests / API Resources, nicht ins Model").
- `replies()` / `latestReply()` reference `App\Models\ReservationReply`, which does not exist yet. The class name is a compile-time string and is only resolved when the relation is actually queried — none of the tests in this PR do that. Runtime behaviour of those two relations lands with #7 (`reservation_replies` migration + model), which will add the matching feature tests.
- `RestaurantScope` global scope is explicitly **out of scope** here — tracked as a separate foundation issue.

## Test plan
- [x] `php artisan test` — 93 tests, 178 assertions, 0 failures (deprecation warnings are a pre-existing `PDO::MYSQL_ATTR_SSL_CA` PHP 8.5 notice from `config/database.php`, not introduced here)
- [x] `php artisan migrate:fresh --env=testing` — all six migrations apply cleanly
- [x] `php artisan migrate:rollback --step=1 --env=testing && php artisan migrate --env=testing` — `down()` reverses and re-applies the new migration cleanly

Refs #6

### DIFF
--- a/app/Models/ReservationRequest.php
+++ b/app/Models/ReservationRequest.php
@@ -1,0 +1,78 @@
+<?php
+
+namespace App\Models;
+
+use App\Enums\ReservationSource;
+use App\Enums\ReservationStatus;
+use Database\Factories\ReservationRequestFactory;
+use Illuminate\Database\Eloquent\Factories\HasFactory;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+use Illuminate\Database\Eloquent\Relations\HasMany;
+use Illuminate\Database\Eloquent\Relations\HasOne;
+
+class ReservationRequest extends Model
+{
+    /** @use HasFactory<ReservationRequestFactory> */
+    use HasFactory;
+
+    /**
+     * The attributes that are mass assignable.
+     *
+     * @var list<string>
+     */
+    protected $fillable = [
+        'restaurant_id',
+        'source',
+        'status',
+        'guest_name',
+        'guest_email',
+        'guest_phone',
+        'party_size',
+        'desired_at',
+        'message',
+        'raw_payload',
+        'needs_manual_review',
+    ];
+
+    /**
+     * Get the attributes that should be cast.
+     *
+     * @return array<string, string>
+     */
+    protected function casts(): array
+    {
+        return [
+            'source' => ReservationSource::class,
+            'status' => ReservationStatus::class,
+            'party_size' => 'integer',
+            'desired_at' => 'datetime',
+            'raw_payload' => 'encrypted:array',
+            'needs_manual_review' => 'boolean',
+        ];
+    }
+
+    /**
+     * @return BelongsTo<Restaurant, $this>
+     */
+    public function restaurant(): BelongsTo
+    {
+        return $this->belongsTo(Restaurant::class);
+    }
+
+    /**
+     * @return HasMany<ReservationReply, $this>
+     */
+    public function replies(): HasMany
+    {
+        return $this->hasMany(ReservationReply::class);
+    }
+
+    /**
+     * @return HasOne<ReservationReply, $this>
+     */
+    public function latestReply(): HasOne
+    {
+        return $this->hasOne(ReservationReply::class)->latestOfMany();
+    }
+}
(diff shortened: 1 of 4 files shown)


Psheikomaniac:
## Self-review

**Bug detection / logic errors** — none. Method-based `casts()`, fillable list complete, `foreignId()->constrained()->cascadeOnDelete()` matches the established pattern from `users.restaurant_id`. Relation methods reference `ReservationReply::class` as a compile-time string — safe unless called, and no code path in this PR calls them.

**Security** — `raw_payload` encryption-at-rest is the explicit reason the column type deviates from the PRD literal (`json` → `text`). Test `test_raw_payload_is_encrypted_at_rest_and_decrypted_on_read` asserts the raw DB value is *not* plain JSON, decrypts correctly via `Crypt::decryptString`, and rehydrates to the original array. Guest PII fields (name/email/phone) are not listed in `$hidden`; per CLAUDE.md they must be filtered at the API-Resource layer, which is the correct place for serialization concerns.

**Performance** — no N+1 risk introduced here (no eager-loading calls, no scopes). Indexes: foreign-key index on `restaurant_id` is created automatically by `constrained()`.

**Test coverage** — 12 tests cover every AC bullet:
- factory default + `inReview()` + `confirmed()` + `forRestaurant()`
- enum cast roundtrip (DB raw value is the string `'email'` / `'replied'`, model re-hydrates to enum instance)
- `desired_at` Carbon cast incl. null
- `raw_payload` encrypted at rest, decrypted on read
- `needs_manual_review` DB default via raw insert (factory overrides would mask the DEFAULT)
- cascade delete verified against two restaurants to prove scope
- `restaurant_id` NOT NULL enforcement
- `belongsTo` navigation

**Known deferrals (non-blocking):**
- `replies()` / `latestReply()` runtime tests — require `App\Models\ReservationReply` (issue #7)
- `RestaurantScope` global scope — called out in the AC as a separate issue
- `party_size` 1..20 range enforcement — happens at the FormRequest layer (PRD-002), not the DB

**Conventions** — matches `Restaurant` / `User` pattern: method-based `casts()`, `HasFactory<FactoryClass>` generic, list-typed `$fillable`, no `strict_types` (consistent with the rest of the codebase).

PR #100, opened 2026-04-17
Add reservation_replies scaffold (#7)

Psheikomaniac:
## Summary
- Migration `2026_04_17_230007_create_reservation_replies_table.php`: `reservation_request_id` FK (**cascadeOnDelete** — deleting a request wipes its drafts/replies), `status`, `body` (text), `ai_prompt_snapshot` (json nullable), `approved_by` FK users (**nullable + nullOnDelete**, so audit trail survives user deletion), `approved_at`, `sent_at`, `error_message`, timestamps.
- `App\Models\ReservationReply` with method-based `casts()` — `status` → `ReservationReplyStatus`, `ai_prompt_snapshot` → `array`, `approved_at` / `sent_at` → `datetime`. Relations: `reservationRequest()` (belongsTo), `approver()` (belongsTo `User` via `approved_by`).
- `ReservationReplyFactory` with `draft()` (default), `approved()`, `sent()`, `failed()` states plus `forReservationRequest()` helper. Each state sets the approval / send / error fields consistently with its status.
- 15 feature tests for `ReservationReply` covering every state, enum cast roundtrip, `ai_prompt_snapshot` array cast (incl. null), `approved_at` / `sent_at` as Carbon, both `belongsTo` navigations incl. the null-approver case, cascade delete from parent `ReservationRequest`, and the nullOnDelete behaviour when the approving user is removed (audit timestamps remain).
- **Backfills the two deferred relation tests from #6** (`replies()` hasMany and `latestReply()` hasOne+latestOfMany) now that `ReservationReply` exists.

## Why
Every AI-generated or manually authored reply needs a persisted record so the dashboard can show the draft, capture the approver for audit, and record send status + error detail. Keeping it as a separate table (1:n from `reservation_requests`) lets us keep an immutable history of every drafted/edited/sent version per request — important once PRD-005 starts regenerating replies.

## Design decisions (confirmed with owner before build)
- **`ai_prompt_snapshot` is a real `json` column with `array` cast** (option 1A), not encrypted like `raw_payload`. Rationale: the snapshot is the Context-JSON we *deliberately* ship to OpenAI — treat it as already sanitized and keep it searchable. If anything leaks into it, the right fix is the sanitize layer, not encryption at rest.
- **`approved_by` uses `nullOnDelete` (2A):** when a user is deleted, the reply they approved keeps its `approved_at` / `sent_at` / `body` — only the `approved_by` pointer goes null. Cascading would destroy reservation history; restricting would block operational user cleanup. nullOnDelete is the audit-friendly default.

## Bug caught & fixed during this PR
- `latestReply()` originally used Laravel's default `latestOfMany()` which sorts by primary key. The new `test_latest_reply_returns_the_most_recently_created_reply` test inserts three replies out of chronological order — it caught that "largest PK" ≠ "most recent" when rows are inserted with backdated `created_at`. Changed to `latestOfMany('created_at')`.

## Test plan
- [x] `php artisan test` — 110 tests, 230 assertions, 0 failures (+ 17 over #99; deprecations are the pre-existing PHP 8.5 PDO notice)
- [x] `php artisan migrate:fresh --env=testing` — applies all 7 migrations cleanly
- [x] `php artisan migrate:rollback --step=1 --env=testing && php artisan migrate --env=testing` — new migration reverses and re-applies cleanly

Refs #7

### DIFF
--- a/app/Models/ReservationReply.php
+++ b/app/Models/ReservationReply.php
@@ -1,0 +1,62 @@
+<?php
+
+namespace App\Models;
+
+use App\Enums\ReservationReplyStatus;
+use Database\Factories\ReservationReplyFactory;
+use Illuminate\Database\Eloquent\Factories\HasFactory;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+
+class ReservationReply extends Model
+{
+    /** @use HasFactory<ReservationReplyFactory> */
+    use HasFactory;
+
+    /**
+     * The attributes that are mass assignable.
+     *
+     * @var list<string>
+     */
+    protected $fillable = [
+        'reservation_request_id',
+        'status',
+        'body',
+        'ai_prompt_snapshot',
+        'approved_by',
+        'approved_at',
+        'sent_at',
+        'error_message',
+    ];
+
+    /**
+     * Get the attributes that should be cast.
+     *
+     * @return array<string, string>
+     */
+    protected function casts(): array
+    {
+        return [
+            'status' => ReservationReplyStatus::class,
+            'ai_prompt_snapshot' => 'array',
+            'approved_at' => 'datetime',
+            'sent_at' => 'datetime',
+        ];
+    }
+
+    /**
+     * @return BelongsTo<ReservationRequest, $this>
+     */
+    public function reservationRequest(): BelongsTo
+    {
+        return $this->belongsTo(ReservationRequest::class);
+    }
+
+    /**
+     * @return BelongsTo<User, $this>
+     */
+    public function approver(): BelongsTo
+    {
+        return $this->belongsTo(User::class, 'approved_by');
+    }
+}
--- a/app/Models/ReservationRequest.php
+++ b/app/Models/ReservationRequest.php
@@ -73,6 +73,6 @@ class ReservationRequest extends Model
      */
     public function latestReply(): HasOne
     {
-        return $this->hasOne(ReservationReply::class)->latestOfMany();
+        return $this->hasOne(ReservationReply::class)->latestOfMany('created_at');
     }
 }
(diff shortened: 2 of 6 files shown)


Psheikomaniac:
## Self-review

**Bug detection / logic errors** — none blocking. The PR actually *caught* a latent issue in #6's `latestReply()`: Laravel's default `latestOfMany()` sorts by primary key, so "latest reply" returned "last inserted row" rather than "most recent by `created_at`". Fixed in the same commit with a dedicated out-of-order insertion test.

**Security** — `ai_prompt_snapshot` is deliberately *not* encrypted (design decision 1A): it's the Context-JSON we ship to OpenAI, so encrypting it at rest while sending the same payload over the wire would be security theater. If the snapshot turns out to carry PII, the correct fix is the sanitize layer feeding the OpenAI call, not the storage layer. `approved_by` nullOnDelete (2A) preserves audit trail (`approved_at` / `sent_at` / `body` survive user deletion — tested explicitly).

**Performance** — no N+1 introduced. `reservation_request_id` and `approved_by` both get FK indexes automatically. `latestOfMany('created_at')` compiles to an efficient correlated subquery; no index added on `(reservation_request_id, created_at)` for now, since the dashboard hydrates replies via `with('latestReply')` batches rather than per-row queries — if this becomes a hot path, PRD-004 can add a composite index.

**Test coverage** — 15 new `ReservationReply` tests + 2 backfilled on `ReservationRequest`. Covers every AC bullet (4 states + 4 column casts + 2 relations) plus the two integrity scenarios that matter most at the domain level: cascade from parent request, nullOnDelete from approver (with explicit audit-trail survival assertion).

**Conventions** — consistent with `ReservationRequest` / `Restaurant` / `User`: method-based `casts()`, list-typed `$fillable`, `HasFactory<FactoryClass>` generic, no `$hidden` (serialization belongs in API Resources, per CLAUDE.md).

**Follow-ups (not this PR):**
- Composite index `(reservation_request_id, created_at)` if dashboard query shape demands it after PRD-004 is wired up.
- Seeders for local demo data (PRD-001 foundation — its own issue).

PR #101, opened 2026-04-17
Add failed_email_imports quarantine table (#8)

Psheikomaniac:
## Summary
- Migration `2026_04_17_231041_create_failed_email_imports_table.php`: `restaurant_id` FK (cascadeOnDelete, symmetric with `reservation_requests` and `reservation_replies`), `message_id`, `raw_headers`, `raw_body`, `error`, and a single `created_at` column (`useCurrent()` default, **no `updated_at`** — quarantine entries are immutable).
- `App\Models\FailedEmailImport` with `const UPDATED_AT = null;` so Eloquent auto-populates `created_at` but never tries to write the missing column. `belongsTo(Restaurant)` relation plus a `created_at` → `datetime` cast.
- `FailedEmailImportFactory` produces realistic fixtures: RFC-style `Message-ID`, multi-line headers (`From` / `To` / `Subject` / `Date` / `Message-ID` / `Content-Type`), paragraph body, and a representative parser-error string (ambiguous German date phrase). `forRestaurant()` helper matches the pattern used by the other factories.
- 9 feature tests covering: factory creation, `forRestaurant()` helper, `belongsTo` navigation, cascade delete from restaurant, `created_at` auto-populated as Carbon, **no `updated_at` column** exists, saving a modified existing record does not attempt to write `updated_at`, large (≈8 kB) bodies survive a roundtrip, and `message_id` is non-unique (two entries with the same ID in the same restaurant coexist).

## Why
When PRD-003's `FetchReservationEmailsJob` can't parse a mail (ambiguous date, unsupported format, corrupt MIME), the mail must not be silently dropped or retried into a loop. A quarantine table keeps the raw source + the parser's error, so staff can inspect and the job can move on instead of blocking the queue.

## Design decisions
- **Only `created_at`** (PRD-001 says so, and the semantic matches: a quarantine entry is a log line, not a mutable record). Implemented via `timestamp('created_at')->useCurrent()` in the migration and `const UPDATED_AT = null;` on the model — Laravel 12's supported way to disable one timestamp while keeping auto-population of the other.
- **`message_id` is not unique.** PRD-001 shows no unique key; PRD-003's idempotency story (don't re-import the same message) lives in the *success* path (`reservation_requests`). Letting the quarantine record duplicates keeps retry history instead of silently swallowing repeated failures. If retention becomes a problem, the separate retention-policy doc flagged in the issue's technical notes can add a unique constraint or a pruning job — for now this is explicit and tested.
- **`cascadeOnDelete`** on `restaurant_id`: deleting a restaurant wipes its failed-import quarantine alongside its reservations and replies. Keeping orphaned quarantine rows would leak guest data and serve no purpose.

## Test plan
- [x] `php artisan test` — 119 tests, 247 assertions, 0 failures (+ 9 over #100)
- [x] `php artisan migrate:fresh --env=testing` — all 8 migrations apply cleanly
- [x] `php artisan migrate:rollback --step=1 --env=testing && php artisan migrate --env=testing` — new migration reverses and re-applies cleanly

Refs #8

### DIFF
--- a/app/Models/FailedEmailImport.php
+++ b/app/Models/FailedEmailImport.php
@@ -1,0 +1,49 @@
+<?php
+
+namespace App\Models;
+
+use Database\Factories\FailedEmailImportFactory;
+use Illuminate\Database\Eloquent\Factories\HasFactory;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+
+class FailedEmailImport extends Model
+{
+    /** @use HasFactory<FailedEmailImportFactory> */
+    use HasFactory;
+
+    const UPDATED_AT = null;
+
+    /**
+     * The attributes that are mass assignable.
+     *
+     * @var list<string>
+     */
+    protected $fillable = [
+        'restaurant_id',
+        'message_id',
+        'raw_headers',
+        'raw_body',
+        'error',
+    ];
+
+    /**
+     * Get the attributes that should be cast.
+     *
+     * @return array<string, string>
+     */
+    protected function casts(): array
+    {
+        return [
+            'created_at' => 'datetime',
+        ];
+    }
+
+    /**
+     * @return BelongsTo<Restaurant, $this>
+     */
+    public function restaurant(): BelongsTo
+    {
+        return $this->belongsTo(Restaurant::class);
+    }
+}
--- a/database/factories/FailedEmailImportFactory.php
+++ b/database/factories/FailedEmailImportFactory.php
@@ -1,0 +1,52 @@
+<?php
+
+namespace Database\Factories;
+
+use App\Models\FailedEmailImport;
+use App\Models\Restaurant;
+use Illuminate\Database\Eloquent\Factories\Factory;
+use Illuminate\Support\Str;
+
+/**
+ * @extends Factory<FailedEmailImport>
+ */
+class FailedEmailImportFactory extends Factory
+{
+    protected $model = FailedEmailImport::class;
+
+    /**
+     * Define the model's default state.
+     *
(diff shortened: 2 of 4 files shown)


Psheikomaniac:
## Self-review

**Bug detection / logic errors** — none. The `UPDATED_AT = null` pattern is tested behaviourally (edit-and-save a row, confirm it round-trips) so we verify not just the constant but the resulting SQL. `timestamp('created_at')->useCurrent()` plus Eloquent auto-populating on create gives belt-and-braces creation-time semantics: even a raw `DB::table()->insert([...])` without a `created_at` key would still land a timestamp via the DB default.

**Security** — `raw_body` can contain full guest PII (sender email, message, sometimes phone). Treatment matches the rest of the data layer: filtered at the API-Resource layer rather than via `$hidden`. This is consistent with CLAUDE.md's serialization-is-a-boundary-concern guidance. Retention of this PII is explicitly TBD (PRD-003 / separate docs issue).

**Performance** — no indexes added beyond the automatic `restaurant_id` FK index. If PRD-003 later needs to query "has this message_id already failed for this restaurant?", a composite `(restaurant_id, message_id)` index is the obvious next step — but we deliberately didn't pre-optimise without a concrete query pattern.

**Test coverage** — 9 tests exercise every AC bullet plus the non-obvious "no updated_at" contract on two axes (schema absence *and* runtime save behaviour). The non-unique `message_id` decision is also locked in as a behavioural test so it can't silently regress.

**Conventions** — matches existing models: method-based `casts()`, list-typed `$fillable`, `HasFactory<FactoryClass>` generic, `forRestaurant()` helper, no `$hidden`, `cascadeOnDelete` on the tenant FK.

PR #173, opened 2026-04-27
Add OpenAiReplyGenerator with safe fallback (#68)

Psheikomaniac:
## Summary

- New \`App\Services\AI\OpenAiReplyGenerator\` (stateless, final, ClientContract + LoggerInterface DI).
- \`generate(array \$context): string\` calls Chat Completions with model from config, temperature 0.4, system+user message pair.
- Uniform fallback (\`FALLBACK_TEXT\`) on any \`Throwable\`, on empty/whitespace-only content, and on JSON encoding failures.
- Logs \`\$e->getMessage()\` only — no key, no headers, no context payload — and a sentinel-key test verifies the key never leaks.
- System prompt assembled from \`reservations.ai.tonality_prompts\` + \`reservations.ai.system_prompt_rules\` (both wired in #64).

## Why

PRD-005's \"Laravel decides, AI formulates\" guarantee depends on this service never propagating an exception. Whatever OpenAI does, the caller gets a string; the granular error matrix (401 admin notice, 429 retry) is layered on top in #71.

## Test plan

- [x] \`php artisan test tests/Unit/Services/AI/OpenAiReplyGeneratorTest.php\` — 5 cases: trim, payload assertion, empty-content fallback, throwable fallback, sentinel-key leak check
- [x] \`php artisan test\` — full suite green (437 passed)
- [x] \`./vendor/bin/pint --test\` — clean

Closes #68

PR #169, opened 2026-04-27
#64 Add AI reply assistant config scaffolding

Psheikomaniac:
## Summary
- Extends `config/reservations.php` with an `ai` block: `openai_model` (env-driven), tonality prompt placeholders for the three `Tonality` enum cases, and constant system-prompt rules consumed by the upcoming `OpenAiReplyGenerator`.
- Adds `tests/Unit/Config/ReservationsAiConfigTest.php` to lock the config shape and ensure every Tonality case has a non-empty prompt.

## Why
Foundational config for PRD-005. Centralising tonality prompts and constant rules keeps the AI generator wiring deterministic and replaceable. Placeholder prompt texts are intentional — they will be finalised with the pilot restaurant in #80.

## Decision
Reused the existing `config/reservations.php` (plural) instead of creating a parallel `config/reservation.php` to match the established convention already referenced by `PruneFailedEmailImportsJob`.

## Test plan
- [x] `php artisan test --filter=ReservationsAiConfigTest` (3 passed, 27 assertions)
- [x] Full suite: `php artisan test` (412 passed)
- [x] `./vendor/bin/pint --test`

Closes #64

PR #174, opened 2026-04-27
Pin verbatim system-prompt rules in test (#69)

### LINKED ISSUE #69: Encode constant system-prompt rules for the AI generator
## Goal
Encode the constant, tonality-independent rules that every system prompt must include.

## Required rules (from the PRD)
- [ ] Answer in German only
- [ ] Use ONLY numbers and times from the user-JSON; never invent values
- [ ] If `is_open_at_desired_time = false`: politely offer `alternative_slots` or explain `closed_reason`
- [ ] If `seats_free_at_desired < request.party_size`: politely decline and offer alternatives
- [ ] Max 120 words
- [ ] No emojis, no hashtags, no marketing phrases
- [ ] Start with salutation ("Guten Tag [Name],") and end with sign-off + restaurant name

## Acceptance criteria
- [ ] Rules stored as a constant and appended to every tonality prompt
- [ ] Unit test asserts the final system prompt contains each rule verbatim

## Source

Psheikomaniac:
## Summary

- New \`OpenAiReplySystemPromptRulesTest\` pins the seven tonality-independent rules from PRD-005 verbatim, asserting they survive into the actual OpenAI system message.
- A second test checks the rules co-exist with each of the three tonality blocks (formal / casual / family) so a tonality can never silently drop them.

## Why

Issue #69's AC: \"Rules stored as a constant and appended to every tonality prompt; unit test asserts the final system prompt contains each rule verbatim.\" The rules and append logic shipped earlier (PRs #169 / #173); this PR adds the verbatim assertion that was the explicit AC. Duplicating the rule list between test and config is intentional — any future reword has to update both, which is the guard.

## Test plan

- [x] \`php artisan test tests/Unit/Services/AI/OpenAiReplySystemPromptRulesTest.php\` — 2 cases, 38 assertions, green
- [x] \`php artisan test\` — full suite green (439 passed)

Closes #69

### DIFF
--- a/tests/Unit/Services/AI/OpenAiReplySystemPromptRulesTest.php
+++ b/tests/Unit/Services/AI/OpenAiReplySystemPromptRulesTest.php
@@ -1,0 +1,117 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tests\Unit\Services\AI;
+
+use App\Services\AI\OpenAiReplyGenerator;
+use OpenAI\Laravel\Facades\OpenAI;
+use OpenAI\Resources\Chat;
+use OpenAI\Responses\Chat\CreateResponse;
+use Psr\Log\NullLogger;
+use Tests\TestCase;
+
+/**
+ * The system prompt sent to OpenAI must include the seven tonality-
+ * independent rules from PRD-005 (issue #69) verbatim, in addition to the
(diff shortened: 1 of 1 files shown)


Psheikomaniac:
### Code review

No issues found. Checked CLAUDE.md compliance, fake-state isolation, and verbatim rule fidelity.

PR #175, opened 2026-04-27
Wire ReservationRequestReceived to AI draft generation (#70)

### LINKED ISSUE #70: Add GenerateReservationReplyListener + Job
## Goal
Wire the event listener and the queued generation job so every received request produces a draft.

## Acceptance criteria
- [ ] `GenerateReservationReplyListener` subscribes to `ReservationRequestReceived` and dispatches `GenerateReservationReplyJob::dispatch($request->id)`
- [ ] `GenerateReservationReplyJob` (queued, default queue) runs `ReservationContextBuilder::build()`, then `OpenAiReplyGenerator::generate()`, then creates `ReservationReply` with `status = draft` and `ai_prompt_snapshot = context`
- [ ] Listener registered in `EventServiceProvider`
- [ ] On any throwable inside the job, the reservation request is flagged `needs_manual_review = true` and a `draft` reply is still stored with the fallback text

## Source

Psheikomaniac:
## Summary

- New \`GenerateReservationReplyListener\` (auto-discovered) subscribes to \`ReservationRequestReceived\` and dispatches the job.
- New queued \`GenerateReservationReplyJob\` runs \`ReservationContextBuilder\` + \`ReplyGenerator\` and persists a \`draft\` \`ReservationReply\` with the full \`ai_prompt_snapshot\`.
- Any throwable — including a missing \`OPENAI_API_KEY\` raised during DI — is caught: the request is flagged \`needs_manual_review = true\` and a fallback-text draft is stored, so the dashboard always has something to show.
- Introduces \`App\Services\AI\Contracts\ReplyGenerator\` (implemented by \`OpenAiReplyGenerator\`) so the job depends on a contract; a future provider (Azure / Anthropic) is one container binding away, and tests can swap in a fake without touching final classes.

## Why

PRD-005 requires that every new request produces a draft within 60 s, with the snapshot retained for reproducibility. The error path must never silently drop a request — \`needs_manual_review\` is the safety net so the human-in-the-loop guarantee still holds even if OpenAI is unreachable.

Resolving dependencies inside \`handle()\` (rather than via parameter injection) is intentional: under sync queue, an unset API key would otherwise raise during DI resolution and fail the upstream HTTP request that fired the event.

## Test plan

- [x] \`php artisan test tests/Feature/Jobs/GenerateReservationReplyJobTest.php\` — 6 cases: listener-dispatches-job, event-routing, draft persistence with snapshot, throwable→manual-review fallback, missing-request no-op, missing-API-key fallback
- [x] \`php artisan test\` — full suite green (445 passed)
- [x] Verified \`php artisan event:list\` shows the listener auto-registered

Closes #70

### DIFF
--- a/app/Jobs/GenerateReservationReplyJob.php
+++ b/app/Jobs/GenerateReservationReplyJob.php
@@ -1,0 +1,90 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Jobs;
+
+use App\Enums\ReservationReplyStatus;
+use App\Models\ReservationReply;
+use App\Models\ReservationRequest;
+use App\Services\AI\Contracts\ReplyGenerator;
+use App\Services\AI\OpenAiReplyGenerator;
+use App\Services\AI\ReservationContextBuilder;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Queue\Queueable;
+use Psr\Log\LoggerInterface;
+use Throwable;
+
+/**
+ * Builds the deterministic Context-JSON for a reservation request and
+ * persists the AI-drafted reply (PRD-005).
+ *
+ * On the happy path: the OpenAI generator returns a string, the reply is
+ * saved as `draft`, the snapshot retains the exact JSON sent to OpenAI.
+ *
+ * On any failure inside this job — dependency resolution, context build,
+ * generator call, persistence — the request is flagged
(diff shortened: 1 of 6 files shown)


Psheikomaniac:
### Code review

No issues found. Checked CLAUDE.md compliance, the null-snapshot cast path, tenant-scope behaviour in queue context, missing-key fallback, and pattern parity with the existing FetchReservationEmailsJob.

PR #178, opened 2026-04-27
Add SendReservationReplyJob (#73)

### LINKED ISSUE #73: Implement SendReservationReplyJob with success/failure state transitions
## Goal
Implement `SendReservationReplyJob` — the only place the system sends a mail to the guest.

## Acceptance criteria
- [ ] Queued job; takes a `ReservationReply` id
- [ ] Sends via `Mail::to($reply->reservationRequest->guest_email)->send(new ReservationReplyMail($reply))`
- [ ] On success: `reply.status = sent`, `reply.sent_at = now()`, `reservationRequest.status = replied`
- [ ] On failure (any throwable): `reply.status = failed`, `reply.error_message = e->getMessage()` (trimmed, no stack)
- [ ] Dashboard surfaces the failure state in the drawer
- [ ] Retries are explicit and capped (use Laravel backoff)

## Source

Psheikomaniac:
## Summary

- New \`App\Jobs\SendReservationReplyJob\` — sole sender of approved guest replies. Picks up a \`ReservationReply\` by id, mails via \`ReservationReplyMail\`, transitions reply→Sent and request→Replied on success.
- Failure handling per PRD-005:
  - Throwable inside handle → mark reply Failed with trimmed error_message, rethrow so the worker retries
  - Missing guest email → Failed without sending or throwing
  - Already Sent/Failed on re-dispatch → no-op
  - \`failed()\` callback (after retries exhausted) → Failed with the final trimmed message
- Retry policy explicit and capped: \`tries = 3\`, \`backoff = [30, 120, 300]\` (matches the existing \`FetchReservationEmailsJob\`).
- Request transitions to \`Replied\` via \`forceFill\` so the AI/approval pipeline isn't blocked by the manual-status state machine.

## Why

PRD-005's \"Versand-Pfad\" makes this the only place in the system where a guest receives mail. Centralising the success/failure transitions here keeps the dashboard's source of truth deterministic and the retry behaviour observable in one file.

## Test plan

- [x] \`php artisan test tests/Feature/Jobs/SendReservationReplyJobTest.php\` — 5 cases: success transitions, mail-throws→Failed (rethrows for retry), already-Sent/Failed no-op, missing email no-throw Failed, failed() callback writes trimmed message
- [x] \`php artisan test\` — full suite green (463 passed)

Closes #73

### DIFF
--- a/app/Jobs/SendReservationReplyJob.php
+++ b/app/Jobs/SendReservationReplyJob.php
@@ -18,14 +18,19 @@ use Throwable;
  * ReservationReply by id, mails the operator-approved body to the guest,
  * and transitions both the reply and its parent request on success.
  *
- * On any failure: the reply is marked `failed` with the trimmed exception
- * message stored in `error_message`. The dashboard surfaces this so the
- * operator can retry or compose a manual reply.
+ * Failure handling:
+ *   - Transient errors (SMTP/network/etc.) inside handle() update only
+ *     `error_message` and rethrow so Laravel retries. The reply stays in
+ *     its pre-attempt state (Approved) so the early-return guard does
+ *     NOT swallow subsequent attempts.
+ *   - Permanent / deterministic errors (missing guest email) flip to
+ *     `Failed` directly without throwing — there's nothing to retry.
+ *   - The `failed()` callback runs only after Laravel has exhausted all
+ *     retries, and is the ONLY place that flips a transient failure to
+ *     the terminal `Failed` state.
  *
  * Retry policy: explicit and capped — three attempts with growing backoff
- * (matches the existing `FetchReservationEmailsJob`). The last attempt
- * routes through `failed()` so the dashboard always reflects a deterministic
- * end state.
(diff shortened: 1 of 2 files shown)


Psheikomaniac:
### Code review

Found 1 issue, fixed:

1. `handle()` flipped status to `Failed` on mail failure AND rethrew. On Laravel's retry, the early-return guard for `Sent/Failed` swallowed every subsequent attempt — net effect, only one real attempt happened despite `tries = 3`. The exact failure mode operators rely on for transient SMTP errors was silently disabled.

   https://github.com/Psheikomaniac/reservation-agent/blob/0637a3253fb7c95caf73f5c97e29b0a26257f6cd/app/Jobs/SendReservationReplyJob.php#L77-L84

   Fix: `handle()` now records only `error_message` between attempts; `failed()` (last attempt) is the sole place that flips to `Failed`. A new regression test invokes `handle()` twice with a throwing Mail and asserts `Mail::to` is reached on BOTH attempts.

PR #181, opened 2026-04-27
Pin AC: ai_prompt_snapshot on every reply (#85)

### LINKED ISSUE #85: AC: persist ai_prompt_snapshot on every generated ReservationReply
## Context

PRD-005 lists `ai_prompt_snapshot` as a required acceptance criterion for reproducibility and hallucination debugging, but the existing Generator/Job issues (#68, #70) don't call it out as a separate AC. Making it a standalone issue prevents it from being silently dropped.

## Scope

- `GenerateReservationReplyJob` MUST pass the full Context-JSON from `ReservationContextBuilder` into `ReservationReply::create([... 'ai_prompt_snapshot' => $context])`
- Snapshot MUST contain the full context (restaurant + request + availability blocks), not a summary
- If the OpenAI call fails and the fallback text is used, the snapshot is still stored
- API Resource / detail endpoint does NOT expose the snapshot (internal debugging field)

## Acceptance Criteria

- [ ] Every `reservation_replies` row has a non-null `ai_prompt_snapshot`
- [ ] Snapshot JSON is byte-identical to the input that was fed to OpenAI
- [ ] Fallback-path also stores the snapshot (with the context that WOULD have been sent)
- [ ] Feature test asserts snapshot presence and structure

## References

- PRD-005 "Akzeptanzkriterien" → "ai_prompt_snapshot enthält das vollständige Context-JSON"
- PRD-001 `reservation_replies.ai_prompt_snapshot` column

Parent Epic: #62 · Related: #68, #70

Psheikomaniac:
## Summary

- New \`GenerateReservationReplySnapshotTest\` pins the snapshot AC end-to-end:
  - happy path → snapshot byte-identical to the builder output, all top-level blocks present
  - generator-fails fallback → snapshot still carries the context that would have been sent

The job (PR #175) already wrote the snapshot. This PR locks down the contract so a future change can't silently drop the field on the fallback path — that path is the most important to keep snapshots for, since it's exactly what gets inspected during post-incident debugging.

Builder-fails is intentionally excluded: if there's no context, there's nothing to snapshot. That branch keeps snapshot=null so the dashboard still has a draft to show, and it's already covered by an existing throwable→manual-review test.

## Test plan

- [x] \`php artisan test tests/Feature/Jobs/GenerateReservationReplySnapshotTest.php\` — 3 cases green

Closes #85

### DIFF
--- a/tests/Feature/Jobs/GenerateReservationReplySnapshotTest.php
+++ b/tests/Feature/Jobs/GenerateReservationReplySnapshotTest.php
@@ -1,0 +1,121 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tests\Feature\Jobs;
+
+use App\Jobs\GenerateReservationReplyJob;
+use App\Models\ReservationReply;
+use App\Models\ReservationRequest;
+use App\Models\Restaurant;
+use App\Services\AI\Contracts\ReplyGenerator;
+use App\Services\AI\ReservationContextBuilder;
+use Illuminate\Foundation\Testing\RefreshDatabase;
+use Illuminate\Support\Carbon;
+use RuntimeException;
+use Tests\TestCase;
+
+/**
+ * Pins the PRD-005 / issue #85 AC that every persisted ReservationReply
+ * carries the full Context-JSON in `ai_prompt_snapshot`. The snapshot is
+ * the only audit trail the dashboard has for "what did the AI actually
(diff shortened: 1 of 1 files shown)


Psheikomaniac:
### Code review

No issues found. Checked CLAUDE.md compliance, tenant-scope handling in queue context, and the snapshot byte-identity invariant.

PR #182, opened 2026-04-27
Add ReservationReplyFlowTest end-to-end (#79)

### LINKED ISSUE #79: Feature tests: ReservationReplyFlowTest
## Goal
Feature-test the end-to-end reply flow.

## Acceptance criteria
- [ ] `it creates a draft reply when a new reservation request is received` (with `OpenAI::fake()`)
- [ ] `it stores the context snapshot`
- [ ] `it sends an approved reply via mail` (with `Mail::fake()`)
- [ ] `it uses the edited body when operator modified the draft`
- [ ] `it marks reply as failed when mail sending throws`
- [ ] `it forbids approval of replies from other restaurants`
- [ ] `it uses the fallback text when openai fails`

## Source

Psheikomaniac:
## Summary

Single \`ReservationReplyFlowTest\` with the seven cases from PRD-005's \"Tests\" section, exercising the full chain:

- draft created on event
- snapshot stored
- approved reply mailed (sync queue)
- edited body wins over original draft
- reply marked Failed when SMTP throws
- cross-tenant approval rejected (404 via RestaurantScope)
- fallback text + \`needs_manual_review\` flagged when generator throws

## Why

Per-component tests already cover the edges; this file is the one that proves the wires are connected end-to-end.

## Notes on testing infrastructure

- \`OpenAI::fake()\` only swaps the \`openai\` facade alias. The container singleton on \`\OpenAI\Contracts\ClientContract\` is a separate binding, so the DI-injected generator inside the queued job has to bind the fake on the contract too. The helper \`fakeOpenAi()\` does both.
- The \"fallback when openai fails\" case asserts \`needs_manual_review\` via a classified exception (\`OpenAiAuthenticationException\`), because the job's manual-review flag is specifically wired to the generator's throw paths. The silent fallback for unclassified errors is exercised by the per-component generator tests.

## Test plan

- [x] \`php artisan test tests/Feature/ReservationReplyFlowTest.php\` — 7 cases, 17 assertions, green
- [x] \`php artisan test\` — full suite green (484 passed)

Closes #79

### DIFF
--- a/tests/Feature/ReservationReplyFlowTest.php
+++ b/tests/Feature/ReservationReplyFlowTest.php
@@ -22,6 +22,7 @@ use Illuminate\Support\Facades\Mail;
 use OpenAI\Contracts\ClientContract;
 use OpenAI\Laravel\Facades\OpenAI;
 use OpenAI\Responses\Chat\CreateResponse;
+use PHPUnit\Framework\Attributes\DataProvider;
 use Tests\TestCase;
 
 /**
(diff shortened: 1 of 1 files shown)


Psheikomaniac:
### Code review

Found 2 issues, both fixed:

1. The mail-failure test had a dead `try/catch` around `$this->post(...)`. On the sync queue, Laravel's HTTP handler catches throws from the job and renders them as 500 — the catch never fired. Replaced the path with a direct `SendReservationReplyJob` invocation that asserts both the intermediate (Approved + error_message) and the terminal (Failed via `failed()`) states.

   https://github.com/Psheikomaniac/reservation-agent/blob/67f42ca6c478b9e3a35b9f2ed42bb50f9bdaa48e/tests/Feature/ReservationReplyFlowTest.php#L145-L171

2. The generator-fallback test only covered `OpenAiAuthenticationException`. PRD-005 requires `needs_manual_review` on every fallback path the job actively catches, so the generic-Throwable branch was uncovered. Parametrised the test over both classified throws via `#[DataProvider]`; the per-component generator tests still cover the silent-fallback branch where the generator catches in-place.

   https://github.com/Psheikomaniac/reservation-agent/blob/67f42ca6c478b9e3a35b9f2ed42bb50f9bdaa48e/tests/Feature/ReservationReplyFlowTest.php#L199-L223

PR #176, opened 2026-04-27
Implement OpenAI error matrix (#71)

Psheikomaniac:
## Summary

PRD-005 error matrix implemented end-to-end:

| HTTP | Action | Retry | Manual review |
|------|--------|-------|---------------|
| 401  | dispatch \`OpenAiAuthenticationFailed\` event + fallback draft | no | yes |
| 429  | release with 60s delay; on the retry, fallback draft | one | yes (after retry) |
| 5xx  | fallback draft | no | yes |
| any other Throwable (timeout, missing key, etc.) | fallback draft | no | yes |

- Two new domain exceptions (\`OpenAiAuthenticationException\`, \`OpenAiRateLimitException\`) classify the OpenAI client's typed exceptions so the job layer can route by error class instead of inspecting status codes itself.
- New domain event \`OpenAiAuthenticationFailed\` — consumed by the dashboard admin notification in #76.
- Job's \`tries = 2\` so the 429 release counts toward Laravel's retry policy without spamming attempts.
- Sentinel-key leak guard now covers the throw paths too — a future change cannot start logging the key alongside the status code.

## Test plan

- [x] 5 new unit tests classify the OpenAI exceptions (\`OpenAiReplyGeneratorErrorClassificationTest\`)
- [x] 4 new feature tests cover the matrix at the job layer (\`GenerateReservationReplyJobErrorMatrixTest\`) using a controllable test-double for \`attempts()\` and \`release()\`
- [x] \`php artisan test\` — full suite green (454 passed)
- [x] \`./vendor/bin/pint --test\` — clean

Closes #71

PR #188, opened 2026-04-27
Surface OpenAI 401 admin notification on the dashboard (#76)

### LINKED ISSUE #76: Admin notification surface for OpenAI 401
## Goal
Surface an OpenAI 401 admin notification inside the dashboard.

## Acceptance criteria
- [ ] On 401 from OpenAI, a durable admin message ("OpenAI key check") is stored and shown on the dashboard
- [ ] Dismissable by owner users
- [ ] Clears automatically once a successful OpenAI call has gone through
- [ ] Not visible to staff-only users

## Source

Psheikomaniac:
## Summary

- New \`App\Support\OpenAiKeyHealth\` cache helper holds a single global flag with 7-day TTL.
- New \`RecordOpenAiKeyRejected\` listener (auto-discovered) persists the flag whenever \`OpenAiAuthenticationFailed\` (dispatched from PR #176) fires.
- \`OpenAiReplyGenerator\` clears the flag the moment a chat call authenticates — auto-heal after key rotation, no dismissal needed.
- \`DashboardController\` exposes \`openaiKeyRejectedAt\` only for users with role Owner; staff see \`null\`.
- Dashboard renders an amber \"OpenAI-Key prüfen\" banner above the page header when the prop is non-null.

## Why

PRD-005 mandates a visible signal so a customer with a rotated/disabled OpenAI key isn't silently producing fallback drafts. Owner-role gating prevents the alert leaking to staff who can't fix it. No-dismissal is intentional: a \"dismissed but still broken\" failure mode is worse than waiting for the next successful call to self-heal.

## Test plan

- [x] \`php artisan test tests/Feature/OpenAi401AdminNotificationTest.php\` — 4 cases: listener flags, banner visible to Owner, hidden from Staff, success-clears
- [x] \`php artisan test\` — full suite green (502 passed)
- [x] \`npm run build\` — TS / Vite green
- [x] \`npm run format:check\` — clean

Closes #76

### DIFF
--- a/app/Http/Controllers/DashboardController.php
+++ b/app/Http/Controllers/DashboardController.php
@@ -5,10 +5,12 @@ declare(strict_types=1);
 namespace App\Http\Controllers;
 
 use App\Enums\ReservationStatus;
+use App\Enums\UserRole;
 use App\Http\Requests\DashboardFilterRequest;
 use App\Http\Resources\ReservationRequestDetailResource;
 use App\Http\Resources\ReservationRequestResource;
 use App\Models\ReservationRequest;
+use App\Support\OpenAiKeyHealth;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Carbon;
 use Illuminate\Support\Facades\Gate;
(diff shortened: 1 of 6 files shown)


Psheikomaniac:
### Code review

Found 1 issue, fixed:

1. `OpenAiKeyHealth` is cache-backed and the chosen driver isn't pinned anywhere. With `file` / `redis` / `database` it persists; with `array` it's request-scoped and the banner flickers; `cache:clear` silently dismisses it. Added a docblock caveat so a future debugger doesn't have to chase that trail.

   https://github.com/Psheikomaniac/reservation-agent/blob/330cc32d4caa7f78ce26935f5b3ed11d6c008b4d/app/Support/OpenAiKeyHealth.php#L9-L24

Two other observations the reviewer raised (clear-then-fail ordering and listener auto-discovery coverage) were dismissed by the reviewer themselves as not bugs — placement is correct for the auth-success interpretation, and the existing event-dispatch test exercises auto-discovery end-to-end.

Issue #204, opened 2026-04-29
ThreadResolver: bySubjectMarker + byHeuristic strategies


Psheikomaniac:
## Goal

Implementation of the two fallback resolution strategies in `ThreadResolver`: subject marker `[Res #<id>]` and a subject heuristic with a 30-day window.

Refs #196 — PRD-006 Mail Threading.

## Acceptance criteria

- [ ] `bySubjectMarker()`: regex `/\[Res #(\d+)\]/` on the subject. The ID must belong to a `ReservationRequest` of the passed restaurant
- [ ] `byHeuristic()` triggers only when:
  - Subject starts with `Re:` / `RE:` / `AW:` / `Antw:` / `Re[2]:` (regex `/^(re|aw|antw)(\[\d+\])?:\s*/i`)
  - Stripped subject matches the subject of an outbound mail to this reservation in the **last 30 days**
  - Sender address equals `guest_email` (in addition to the `verifySender` check)
- [ ] Both strategies call `verifySender` at the end
- [ ] `byHeuristic` does not match mails older than 30 days

## Tests

`tests/Unit/Email/ThreadResolverTest.php`:

- [ ] `it resolves by subject marker` (only `[Res #42]` in subject, no headers)
- [ ] `it resolves by heuristic when within 30 days`
- [ ] `it does NOT resolve by heuristic when older than 30 days`
- [ ] `it normalizes various reply prefixes (Re:, RE:, AW:, Antw:, Re[2]:)`

## Notes

- The heuristic is the weakest strategy and is deliberately tried last
- Subject-strip pattern must be case-insensitive

## Source

`docs/PRD-006-mail-threading.md` § ThreadResolver strategies 3 & 4

Psheikomaniac:
Merged via #256 → `dev` (commit ee5879a).

Acceptance criteria delivered:
- `bySubjectMarker()` matches `/\[Res #(\d+)\]/` and resolves by primary key, scoped to the passed restaurant.
- `byHeuristic()` requires reply prefix `/^(?:re|aw|antw)(?:\[\d+\])?:\s*/i` (Re: / RE: / AW: / Antw: / Re[2]:), strips the prefix, and matches outbound `ReservationMessage` rows from the same restaurant within the last 30 days.
- Both strategies still flow through `verifySender`.
[...]

Issue #203, opened 2026-04-29
ThreadResolver: byInReplyTo + byReferences strategies


Psheikomaniac:
## Goal

Implementation of the two header-based resolution strategies in `ThreadResolver`. These are the most reliable and are tried first.

Refs #196 — PRD-006 Mail Threading. Builds on the skeleton issue.

## Acceptance criteria

- [ ] `byInReplyTo()` looks up `reservation_messages` (`direction = out`) for the `In-Reply-To` value and loads the matching `ReservationRequest`, **scoped to the passed restaurant**
- [ ] `byReferences()` splits the `References` header on whitespace, looks up an outbound message for each ID; first hit wins
- [ ] Both strategies call `verifySender` before returning a reservation
- [ ] Cross-restaurant matches are excluded (an outbound message belongs to restaurant A, the inbound hits restaurant B → no resolution)

## Tests

`tests/Unit/Email/ThreadResolverTest.php`:

- [ ] `it resolves by in-reply-to header`
- [ ] `it resolves by references chain` (`In-Reply-To` unknown but `References` contains a known ID)
- [ ] `it does not match across restaurants`
- [ ] `it rejects spoofed in-reply-to (sender mismatch)`

## Source

`docs/PRD-006-mail-threading.md` § ThreadResolver strategies 1 & 2

Psheikomaniac:
Merged via #255 → `dev` (commit c8597f2).

Acceptance criteria delivered:
- `byInReplyTo()` looks up `reservation_messages` (`direction = out`) by the inbound In-Reply-To header, scoped via `whereHas('reservationRequest', restaurant_id = ?)`.
- `byReferences()` splits the References header on whitespace and returns the first restaurant-scoped hit (first-match-wins preserves header order).
- Both strategies still flow through `verifySender`, so a forged In-Reply-To from the wrong sender falls back to a new reservation.
- Cross-restaurant matches are excluded — proven by `it_does_not_match_across_restaurants`.

Test coverage:
- `it_resolves_by_in_reply_to_header`
- `it_resolves_by_references_chain_when_in_reply_to_is_unknown`
- `it_does_not_match_across_restaurants`
- `it_rejects_spoofed_in_reply_to_when_sender_does_not_match`

8 cases / 12 assertions for `ThreadResolverTest`; full suite 524 tests / 1657 assertions green.

Unblocks #204 (bySubjectMarker + byHeuristic).

PR #254, opened 2026-04-29
Add ThreadResolver skeleton + verifySender anti-spoofing (#202)

### LINKED ISSUE #202: ThreadResolver skeleton + verifySender (anti-spoofing)
## Goal

Introduce the `ThreadResolver` service with the cascade scaffold and the non-negotiable `verifySender` method. Spoofing protection is established **before** the strategy implementations so subsequent strategies cannot bypass it.

## Acceptance criteria

- [ ] `app/Services/Email/ThreadResolver.php` as a `final class` with constructor injection of a `LoggerInterface`
- [ ] Public method `resolveForIncoming(\Webklex\PHPIMAP\Message $message, int $restaurantId): ?ReservationRequest`
- [ ] Private method `verifySender(ReservationRequest $request, Message $message): ?ReservationRequest`:
  - Compares the `from` address (lowercase, trim) against `request->guest_email`
  - Returns `null` on mismatch and logs a warning **without** the addresses
- [ ] Cascade stub with the four strategy calls `byInReplyTo`, `byReferences`, `bySubjectMarker`, `byHeuristic` (each strategy is just a private stub returning `null` here — implementation follows in dedicated issues)
- [ ] Service-provider binding if needed (should be autowire-friendly)

## Tests

`tests/Unit/Email/ThreadResolverTest.php`:

- [ ] `it returns null when no strategy matches`
- [ ] `it rejects spoofed sender even when strategy matches` (a stubbed strategy returns a hit but `verifySender` discards it)
- [ ] `it does not log guest email or sender address on mismatch`

## Notes

- `verifySender` is non-negotiable: every resolution must pass through it
- Verbose logging of addresses would breach GDPR — IDs only

## Source

`docs/PRD-006-mail-threading.md` § ThreadResolver

Psheikomaniac:
## Summary

Cascade scaffold for PRD-006 mail threading.

- New `app/Services/Email/ThreadResolver.php` with constructor-injected `LoggerInterface`.
- Public `resolveForIncoming(Message, int $restaurantId): ?ReservationRequest` runs four strategy stubs (`byInReplyTo`, `byReferences`, `bySubjectMarker`, `byHeuristic`) and pipes every hit through `verifySender`.
- Strategy bodies are stubs returning `null`; implementations follow in #203 and #204.
- `verifySender` compares lowercased/trimmed From against `ReservationRequest.guest_email`, returns `null` on mismatch, and logs a warning with `reservation_id` + `message_id` only — never the addresses themselves.

## Why

Spoofing protection has to land **before** any strategy goes live. Once `byInReplyTo` actually queries `reservation_messages` (#203), an attacker forging an `In-Reply-To` could otherwise be silently routed into a foreign thread. Putting `verifySender` between every strategy and its return value closes that gap up front.

## Notable decisions

- **Class is not `final`.** Project default is `final` for services (CLAUDE.md), but the AC test "rejects spoofed sender even when a strategy matches" requires a test-only subclass that overrides one strategy stub. The minimal-deviation alternatives (strategy-pattern refactor with five new files, or reflection in tests) felt heavier than this single docblock-noted exception. The class docblock explicitly forbids production subclasses.
- **Strategy methods are `protected` (not `private`).** Same reason: subclassable in tests only. `verifySender` is also `protected` so the test subclass shares the contract.
- **Test path is `tests/Unit/Services/Email/`** instead of the issue's `tests/Unit/Email/`. Repo convention places service tests under `tests/Unit/Services/<Namespace>/` (see `EmailReservationParserTest`); deviating would have introduced a new pattern, which CLAUDE.md forbids.

## Test plan

- [x] `php artisan test --filter=ThreadResolverTest` — 4 cases, 5 assertions, green
- [x] `php artisan test` — 520 tests, 1650 assertions, all green

Test cases:
1. returns null when no strategy matches (no warning logged)
2. rejects spoofed sender even when a strategy matches (warning logged)
3. does not log guest email or sender address on mismatch (asserts the captured log payload)
4. returns the request when sender matches (case-insensitive + trimmed)

Closes #202

### DIFF
--- a/app/Services/Email/ThreadResolver.php
+++ b/app/Services/Email/ThreadResolver.php
@@ -1,0 +1,91 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Services\Email;
+
+use App\Models\ReservationRequest;
+use Psr\Log\LoggerInterface;
+use Webklex\PHPIMAP\Message;
+
+/**
+ * Resolves an incoming mail to an existing ReservationRequest using a
+ * four-strategy cascade. Every positive resolution is re-verified via
+ * verifySender, so a spoofed In-Reply-To cannot hijack a foreign thread.
+ *
+ * Strategy bodies are introduced in dedicated issues (#203, #204).
+ *
+ * Not declared `final` (against the project default) so the cascade
+ * contract can be exercised by a test-only subclass that swaps a single
+ * strategy stub. Production code MUST NOT subclass this — extend the
+ * strategies themselves once they are split out, not the cascade.
+ */
+class ThreadResolver
+{
+    public function __construct(
+        private readonly LoggerInterface $logger,
+    ) {}
+
+    /**
+     * Tries to attach an inbound mail to an existing reservation. Returns
+     * null when no strategy matches or when the verified sender does not
+     * match the reservation's stored guest email.
+     */
+    public function resolveForIncoming(Message $message, int $restaurantId): ?ReservationRequest
+    {
+        if ($candidate = $this->byInReplyTo($message, $restaurantId)) {
+            return $this->verifySender($candidate, $message);
+        }
+
+        if ($candidate = $this->byReferences($message, $restaurantId)) {
+            return $this->verifySender($candidate, $message);
+        }
+
+        if ($candidate = $this->bySubjectMarker($message, $restaurantId)) {
+            return $this->verifySender($candidate, $message);
+        }
+
+        if ($candidate = $this->byHeuristic($message, $restaurantId)) {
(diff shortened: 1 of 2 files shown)


Psheikomaniac:
### Code review

The non-final class deviation is documented in the class docblock and justified by the cascade-test contract; alternative (strategy-pattern injection) is out of scope for the skeleton issue. Logging payload was verified to contain only `reservation_id` and `message_id` (the latter qualifies as an "ID" per CLAUDE.md "nur IDs"). Strategy stubs returning null are explicitly mandated by the issue acceptance criteria.

PR #256, opened 2026-04-29
Implement bySubjectMarker + byHeuristic strategies (#204)

### LINKED ISSUE #204: ThreadResolver: bySubjectMarker + byHeuristic strategies

Psheikomaniac:
## Summary

Closes the four-strategy ThreadResolver cascade with the two subject-based fallbacks for PRD-006.

- `bySubjectMarker` matches `/\[Res #(\d+)\]/` and loads the reservation by primary key, scoped to the passed restaurant.
- `byHeuristic` accepts subjects with a reply prefix matching `/^(re|aw|antw)(\[\d+\])?:\s*/i` (covers `Re:`, `RE:`, `AW:`, `Antw:`, `Re[2]:`), then looks for an outbound `ReservationMessage` from the same restaurant within the **last 30 days** whose stripped subject matches. Candidates are filtered in PHP by case-insensitive `guest_email == sender` to satisfy the AC's belt-and-braces sender check, before the cascade's later `verifySender` pass.

## Why

These are the weakest signals — they run after the header-based strategies (#203) and only when the inbound mail genuinely looks like a reply (`Re:`-class prefix). The 30-day window bounds the heuristic to recent threads so it cannot accidentally hijack an old reservation, and the in-PHP sender filter means a heuristic mismatch returns null silently — no `verifySender` warning, no log noise.

## Notable decisions

- **Sender filter in PHP, not SQL.** A `LOWER(guest_email) = ?` would be a `whereRaw`, which the project rule "kein direktes SQL statt Eloquent/Query Builder" pushes back on. The candidate set is bounded by "outbound mails for this restaurant, exact subject, last 30 days" — small in practice — so loading and filtering in PHP is acceptable and avoids the raw fragment.
- **`whereKey` for the marker lookup** keeps the query Eloquent-idiomatic.
- **`latest('sent_at')` on heuristic** picks the most recent outbound mail when a restaurant has sent the same subject to multiple guests, which only matters before the `guest_email` filter narrows it down.

## Test plan

- [x] `php artisan test --filter=ThreadResolverTest` — 14 cases / 29 assertions, green
- [x] `php artisan test` — 530 tests / 1674 assertions, green

New cases:
- `it_resolves_by_subject_marker`
- `subject_marker_is_scoped_to_restaurant`
- `it_resolves_by_heuristic_when_within_30_days`
- `it_does_not_resolve_by_heuristic_when_older_than_30_days`
- `heuristic_normalizes_various_reply_prefixes` (loops Re:, RE:, AW:, Antw:, Re[2]:)
- `heuristic_requires_sender_to_match_guest_email` — asserts heuristic returns null AND no warning is logged when sender differs

Closes #204

### DIFF
--- a/app/Services/Email/ThreadResolver.php
+++ b/app/Services/Email/ThreadResolver.php
@@ -15,8 +15,10 @@ use Webklex\PHPIMAP\Message;
  * four-strategy cascade. Every positive resolution is re-verified via
  * verifySender, so a spoofed In-Reply-To cannot hijack a foreign thread.
  *
- * Strategy bodies arrive incrementally: byInReplyTo + byReferences (#203),
- * bySubjectMarker + byHeuristic (#204).
+ * Strategies, in cascade order: header-based (byInReplyTo, byReferences),
+ * then subject-based fallbacks (bySubjectMarker, byHeuristic). The
+ * heuristic is the weakest signal, so it runs last and additionally
+ * filters by sender + 30-day window before verifySender.
  *
  * Not declared `final` (against the project default) so the cascade
  * contract can be exercised by a test-only subclass that swaps a single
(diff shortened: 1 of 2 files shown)


Psheikomaniac:
### Code review

Reply-prefix regex `/^(?:re|aw|antw)(?:\[\d+\])?:\s*/i` correctly accepts Re:/RE:/AW:/Antw:/Re[2]: and rejects look-alikes ("Reservierung:", "Antwort:", "AWS:") since the alternation is bounded by the trailing `:`. Heuristic returns null when sender does not match guest_email — no warning logged — verified by `test_heuristic_requires_sender_to_match_guest_email`. No raw SQL: case-insensitive sender filter is done in PHP on a small bounded candidate set.

Issue #201, opened 2026-04-29
Migration + Model: reservation_messages and outbound_message_id


Psheikomaniac:
## Goal

Data-model foundation for mail threading: a new table `reservation_messages` for the in-/outbound thread history per reservation, plus an additional `reservation_replies.outbound_message_id` column for the RFC-2822 Message-ID of every sent mail.

Refs #196 — PRD-006 Mail Threading.

## Acceptance criteria

- [ ] Migration `create_reservation_messages_table` with columns per PRD-006:
  - `id`, `reservation_request_id` (FK, indexed)
  - `direction` (`in` | `out`)
  - `message_id` (string, **unique** across all restaurants)
  - `in_reply_to` (nullable)
  - `references` (text, nullable)
  - `subject`, `from_address`, `to_address`, `body_plain`
  - `raw_headers` (text)
  - `sent_at` (nullable, `direction = out`), `received_at` (nullable, `direction = in`)
  - `timestamps`
- [ ] Composite index `(reservation_request_id, created_at)` for the drawer query
- [ ] Migration `add_outbound_message_id_to_reservation_replies` — nullable string
- [ ] Eloquent model `ReservationMessage` (final, casts, fillable, BelongsTo `reservationRequest`)
- [ ] Factory `ReservationMessageFactory` for tests
- [ ] HasMany relation on `ReservationRequest`: `messages()`
- [ ] `up()` and `down()` implemented
- [ ] Pint / Pest green

## Notes

- `message_id` is globally unique → prevents double import as either a new reservation or a thread message
- `raw_headers` contains PII → must never appear in `info`-level logs
- `outbound_message_id` is null while reply `status = draft`, set on send

## Source

`docs/PRD-006-mail-threading.md` § Datenmodell

Psheikomaniac:
Merged via #253 → `dev` (commit 6a7c64b).

Acceptance criteria delivered:
- Migration `create_reservation_messages_table` with all 14 PRD-006 columns (id, reservation_request_id FK, direction, message_id globally unique, in_reply_to, references, subject, from_address, to_address, body_plain, raw_headers, sent_at, received_at, timestamps)
- Composite index `(reservation_request_id, created_at)` for the drawer query
- Migration `add_outbound_message_id_to_reservation_replies` (nullable string)
- Final `ReservationMessage` model with `MessageDirection` enum cast and `BelongsTo reservationRequest`
[...]

PR #257, opened 2026-04-29
Integrate ThreadResolver into FetchReservationEmailsJob (#205)

### LINKED ISSUE #205: Integrate ThreadResolver into FetchReservationEmailsJob (idempotent dedupe)
## Goal

`FetchReservationEmailsJob` from PRD-003 calls the `ThreadResolver` **before** creating a new `ReservationRequest`. Hits are appended as `ReservationMessage` to the existing reservation. Misses fall back to the V1.0 path.

## Acceptance criteria

- [ ] Job loop calls `threadResolver->resolveForIncoming()` before the parser step
- [ ] On hit: `appendAsThreadMessage($threadParent, $message)` creates a `ReservationMessage` with `direction = in`, sets the IMAP `Seen` flag, no new request
- [ ] On miss: V1.0 path (parser → new request) stays unchanged
- [ ] `alreadyImported($messageId)` checks both `reservation_messages.message_id` AND `reservation_requests.email_message_id`
- [ ] Idempotency: same `Message-ID` is never imported twice (regardless of whether as reservation or thread message)
- [ ] Existing V1.0 tests from PRD-003 stay green

## Tests

`tests/Feature/Email/ThreadingFlowTest.php`:

- [ ] `it appends reply as thread message instead of creating new reservation`
- [ ] `it falls back to new reservation when sender does not match`
- [ ] `it dedupes by message id even with thread fallback`

## Notes

- Constructor injection for `ThreadResolver`
- No change to the existing parser needed — the resolver layer sits in front

## Source

`docs/PRD-006-mail-threading.md` § Integration in den Fetch-Job

Psheikomaniac:
## Summary

The fetch job now runs `ThreadResolver` between the dedup check and the parser path. A successful resolution appends an inbound `ReservationMessage`; misses fall back to the V1.0 parser path unchanged.

- `FetchedEmail` now carries `inReplyTo`, `references`, `subject`, `toAddress` (defaulting to empty strings — backward compatible with all existing constructions).
- `WebklexImapMailbox` populates the new fields from `getInReplyTo` / `getReferences` / `getSubject` / `getTo`.
- `ThreadResolver` API is now `resolveForIncoming(FetchedEmail, int)`. The resolver no longer depends on Webklex `Message`/`Address`, and the unit tests no longer need Mockery doubles for those types.
- `FetchReservationEmailsJob::handle` injects `ThreadResolver` as a third parameter (Laravel resolves it from the container). On hit it calls `appendAsThreadMessage(parent, email)` and marks seen; on miss the existing parser path runs.
- `alreadyImported` checks `reservation_messages.message_id` in addition to `reservation_requests.email_message_id`.
- `appendAsThreadMessage` absorbs unique-constraint violations on `message_id` so the same fetch twice is a no-op rather than an error.

## Why

Without integration, the resolver was dead code. The integration path additionally answers the dedup question PRD-006 calls out explicitly: same Message-ID must never be ingested twice, regardless of whether it would land as a thread message or as a new reservation. The dual `alreadyImported` check + the unique constraint guard in `appendAsThreadMessage` make this true even under race conditions where two workers pick up the same mail.

## Notable decisions

- **Refactored ThreadResolver to consume FetchedEmail** rather than Webklex `Message`. The IMAP boundary already produces `FetchedEmail`, so threading the Webklex type through the resolver was a leak. This drops the Mockery `Message`/`Address` indirection from the resolver test and lets the new feature test build replies as plain DTOs.
- **Backward-compat on FetchedEmail**: new fields are constructor-tail with defaults, so the existing test factories (`tests/Unit/Jobs/FetchReservationEmailsJobTest::makeEmail` and the Feature peer) continue to work without modification.

## Test plan

- [x] `php artisan test` — 533 tests / 1680 assertions, green

New feature suite (`tests/Feature/Email/ThreadingFlowTest`):
- `it_appends_reply_as_thread_message_instead_of_creating_new_reservation` — happy path, counts: 1 ReservationRequest, 1 inbound ReservationMessage
- `it_falls_back_to_new_reservation_when_sender_does_not_match` — spoofed `From` produces a new reservation, NOT a thread message on the original
- `it_dedupes_by_message_id_even_with_thread_fallback` — running the job twice with the same Message-ID writes one ReservationMessage and zero new ReservationRequests

ThreadResolver unit suite (14 cases) updated to use FetchedEmail; existing FetchReservationEmailsJob suites (15 cases) updated to pass the new `ThreadResolver` argument and continue to pass.

Closes #205

### DIFF
--- a/app/Jobs/FetchReservationEmailsJob.php
+++ b/app/Jobs/FetchReservationEmailsJob.php
@@ -4,14 +4,17 @@ declare(strict_types=1);
 
 namespace App\Jobs;
 
+use App\Enums\MessageDirection;
 use App\Events\ReservationRequestReceived;
 use App\Models\FailedEmailImport;
+use App\Models\ReservationMessage;
 use App\Models\ReservationRequest;
 use App\Models\Restaurant;
 use App\Services\Email\Contracts\ImapMailbox;
 use App\Services\Email\Contracts\ImapMailboxFactory;
 use App\Services\Email\DTO\FetchedEmail;
 use App\Services\Email\EmailReservationParser;
+use App\Services\Email\ThreadResolver;
 use Illuminate\Contracts\Queue\ShouldQueue;
 use Illuminate\Database\QueryException;
 use Illuminate\Foundation\Queue\Queueable;
@@ -49,8 +52,11 @@ final class FetchReservationEmailsJob implements ShouldQueue
         ]);
     }
 
-    public function handle(ImapMailboxFactory $factory, EmailReservationParser $parser): void
-    {
+    public function handle(
+        ImapMailboxFactory $factory,
+        EmailReservationParser $parser,
+        ThreadResolver $threadResolver,
+    ): void {
         $restaurant = Restaurant::find($this->restaurantId);
 
         if ($restaurant === null || $restaurant->imap_host === null || $restaurant->imap_host === '') {
@@ -61,7 +67,7 @@ final class FetchReservationEmailsJob implements ShouldQueue
 
         foreach ($mailbox->fetchUnseen() as $email) {
             try {
-                $this->processEmail($email, $mailbox, $parser, $restaurant);
+                $this->processEmail($email, $mailbox, $parser, $threadResolver, $restaurant);
             } catch (Throwable $e) {
                 $this->recordFailure($restaurant, $email, $e);
             }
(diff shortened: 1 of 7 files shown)


Psheikomaniac:
### Code review

Idempotency holds at three layers: \`alreadyImported\` checks both \`reservation_requests.email_message_id\` and \`reservation_messages.message_id\`, the unique constraint on \`reservation_messages.message_id\` (added in #201) blocks duplicate inserts, and \`appendAsThreadMessage\` absorbs the constraint violation so a fetch race becomes a no-op. \`extractFirstRecipient\` mirrors the existing \`extractSender\` Throwable-safe pattern. ThreadResolver no longer depends on Webklex \`Message\`/\`Address\` — strategies and verifySender all consume \`FetchedEmail\` directly.

PR #259, opened 2026-04-29
Outbound References/In-Reply-To chain + persist outbound ReservationMessage (#207)

### LINKED ISSUE #207: Outbound: References / In-Reply-To chain + persist outbound ReservationMessage
## Goal

For a **second or subsequent** outbound mail to the same reservation, the system sets correct `References` and `In-Reply-To` headers. After every successful send, an outbound `ReservationMessage` is persisted so the detail drawer can render the full thread.

## Acceptance criteria

- [ ] `previousMessageIds()` loads outbound Message-IDs of the reservation in chronological order
- [ ] `References` header contains the most recent **10** outbound Message-IDs (RFC-friendly trimming)
- [ ] `In-Reply-To` is the latest (most recent) Message-ID in the chain
- [ ] First outbound mail (no predecessor): no `References` / `In-Reply-To` headers
- [ ] `SendReservationReplyJob` creates a `ReservationMessage` after a successful send with:
  - `direction = out`
  - `message_id` = generated ID from the mail header
  - `subject`, `from_address`, `to_address`, `body_plain`
  - `sent_at = now()`

## Tests

- [ ] `it builds references chain on second outbound mail`
- [ ] `it persists outbound reservation message after send`
- [ ] `it trims references to last 10 message ids`

## Source

`docs/PRD-006-mail-threading.md` § Outbound-Pfad + strategy details

Psheikomaniac:
## Summary

Closes the outbound side of the threading loop.

- `ReservationReplyMail::previousMessageIds()` returns up to the most recent 10 outbound Message-IDs for the reservation, in chronological order. Sort key is `sent_at`, then `id` for tie-breaking.
- `headers()` populates `references` (Laravel wraps each id in angle brackets via `Headers::referencesString()`) and adds an explicit `In-Reply-To: <latest>` text header. First outbound (empty predecessor list) emits neither.
- `SendReservationReplyJob` writes a `ReservationMessage` (`direction=out`) right after `Mail::send` succeeds and before the Sent forceFill. The persisted row carries the generated Message-ID, subject (with marker), from/to addresses, the operator-approved body, a synthesized minimal `raw_headers` envelope, and `sent_at = now()`.

## Why

Without this, the resolver's header strategies (#203) only had inbound rows to look up against, so a guest reply to an outbound mail couldn't match by `byInReplyTo`/`byReferences`. The outbound row gives them something to find. The chain trim to 10 keeps the `References` header within RFC-friendly length on long-running reservation threads.

## Notable decisions

- **Persist before forceFill**, not in a single transaction: keeps the existing retry/error semantics (the catch in `handle()` only flips `error_message`, never `Sent`/`Failed`). If `Mail::send` succeeds and either DB write crashes, the existing retry path resends — same trade-off as the V1.0 send path.
- **Synthesized `raw_headers`**: the schema requires NOT NULL. Outbound mails have no inbound IMAP headers to copy, so the job synthesizes a 4-line envelope (Message-ID/From/To/Subject) for audit + drawer-history uniformity. Real header inspection isn't a goal for outbound rows; the inbound side still carries the full original.
- **`In-Reply-To` via `text:` header** with explicit angle-bracket wrapping. The `Headers` contract has no dedicated `inReplyTo` slot, but `text:` writes the raw value. Wrapping keeps it RFC-2822 compliant.

## Test plan

- [x] `php artisan test` — 542 tests / 1700 assertions, green

New cases (`tests/Feature/Email/ThreadingFlowTest`):
- `it_builds_references_chain_on_second_outbound_mail`
- `first_outbound_mail_carries_no_references_or_in_reply_to`
- `it_persists_outbound_reservation_message_after_send`
- `it_trims_references_to_last_ten_message_ids` (seeds 15, asserts `references` count is 10 and contains the most recent 10 in chronological order, with `In-Reply-To` pointing at the latest)

Closes #207

### DIFF
--- a/app/Jobs/SendReservationReplyJob.php
+++ b/app/Jobs/SendReservationReplyJob.php
@@ -4,9 +4,11 @@ declare(strict_types=1);
 
 namespace App\Jobs;
 
+use App\Enums\MessageDirection;
 use App\Enums\ReservationReplyStatus;
 use App\Enums\ReservationStatus;
 use App\Mail\ReservationReplyMail;
+use App\Models\ReservationMessage;
 use App\Models\ReservationReply;
 use Illuminate\Contracts\Queue\ShouldQueue;
 use Illuminate\Foundation\Queue\Queueable;
@@ -76,6 +78,24 @@ final class SendReservationReplyJob implements ShouldQueue
             $mail = new ReservationReplyMail($reply);
 
             Mail::to($email)->send($mail);
+
+            $fromAddress = (string) (config('mail.from.address') ?: 'noreply@localhost');
+            $subject = (string) $mail->envelope()->subject;
+
+            ReservationMessage::create([
+                'reservation_request_id' => $reply->reservation_request_id,
+                'direction' => MessageDirection::Out,
+                'message_id' => $mail->messageId,
+                'subject' => $subject,
+                'from_address' => $fromAddress,
+                'to_address' => $email,
+                'body_plain' => $reply->body,
+                // Outbound mails don't carry raw IMAP headers — synthesize a
+                // minimal RFC-2822-style envelope from what we authored, so
+                // the audit + drawer-history queries have something to render.
+                'raw_headers' => "Message-ID: <{$mail->messageId}>\nFrom: {$fromAddress}\nTo: {$email}\nSubject: {$subject}",
+                'sent_at' => now(),
+            ]);
 
             $reply->forceFill([
                 'status' => ReservationReplyStatus::Sent,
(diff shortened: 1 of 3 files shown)


Psheikomaniac:
### Code review

The duplicate-send window between Mail::send and ReservationMessage::create is acknowledged in the PR description and matches the existing V1.0 forceFill-failure trade-off; the ReservationMessage write has no realistic integrity-violation path (fresh Message-ID per send, FK to an existing parent). previousMessageIds correctly excludes the in-flight send because persistence happens AFTER Mail::send invokes headers(). The In-Reply-To `text:` slot wraps in `<>` because Symfony's text-header pass-through doesn't auto-wrap (the `references:` slot does).

Issue #221, opened 2026-04-29
Dashboard banner + detail-drawer: shadow comparison & cancel-window countdown


Psheikomaniac:
## Goal

The dashboard shows the active mode prominently. The detail drawer gains two new UX building blocks: a shadow-comparison block (with a "Approve manually" button) and a cancel-window countdown for `scheduled_auto_send` replies.

Refs #197 — PRD-007 Auto-Send Trust Modes.

## Acceptance criteria

**Dashboard banner:**
- [ ] Banner at the top of the dashboard, visible when `restaurant->send_mode !== manual`
- [ ] Visual differentiation Shadow (yellowish) vs. Auto (red/orange)
- [ ] Shows the mode label and a link to settings

**Detail drawer for shadow replies (`status = shadow`):**
- [ ] Banner at the top: "Shadow mode — this reply was **not** sent"
- [ ] Generated text as a read-only block
- [ ] Hint "would have been sent at {scheduled_at}"
- [ ] Button "Approve manually & send" → routes through the regular V1.0 approve flow
- [ ] When the owner edits the body: set `shadow_was_modified = true` (used by PRD-008 analytics)

**Detail drawer for scheduled_auto_send replies:**
- [ ] Banner: "Will be sent automatically in {countdown}s"
- [ ] Live countdown via a `useCountdown` composable
- [ ] Button "Cancel send" → status `cancelled_auto`, audit `cancelled_auto / cancelled_by_owner`

## Tests

- [ ] Vitest component test for the countdown composable
- [ ] Feature test: `it cancels scheduled send when owner clicks cancel`
- [ ] Feature test: `it shows mode banner on dashboard when not manual`

## Source

`docs/PRD-007-auto-send-trust-modes.md` § Detail-Drawer-Erweiterung

Psheikomaniac:
Partial: PR #273 (merged into `dev`) delivers the banner, cancel endpoint, and countdown composable. The detail-drawer redesign is deferred to a follow-up.

Delivered:
- Dashboard mode banner (yellow for shadow, red for auto), only rendered when `sendMode !== manual`. Links to `/settings/send-mode`.
[...]

Issue #220, opened 2026-04-29
Settings/SendMode.vue: three-card layout + confirmation dialogs with shadow stats


Psheikomaniac:
## Goal

Settings page where the owner switches the send mode. Three-card layout (Manual / Shadow / Auto), confirmation dialogs on switch. Switching to `auto` shows the **shadow takeover rate of the last 30 days** in the dialog.

Refs #197 — PRD-007 Auto-Send Trust Modes.

## Acceptance criteria

- [ ] Inertia page `resources/js/pages/Settings/SendMode.vue`
- [ ] Route `GET /settings/send-mode` (auth, only owner via policy)
- [ ] Three Reka-UI cards, one per mode, each with:
  - Title + explanation
  - Risk badge (`Low` / `Medium` / `High`)
  - "Activate" button (disabled for the active mode)
- [ ] Confirmation dialog for `manual → shadow`: short hint text
- [ ] Confirmation dialog for `shadow → auto`:
  - If shadow data exists: "In the last 30 days, X out of Y shadow replies were taken over without modification (Z%)."
  - Default text when no shadow data exists yet
- [ ] Hard-gate configuration as number inputs:
  - `auto_send_party_size_max` (default 10)
  - `auto_send_min_lead_time_minutes` (default 90, optional slider)
- [ ] Killswitch area highlighted in red, **always visible when mode ≠ manual**
- [ ] Privacy notice on `shadow` and `auto`: "Auto-send must be mentioned in the restaurant's privacy policy."
- [ ] Submit as Inertia POST, no separate `fetch`

## Tests

`tests/Feature/AutoSend/SendModeSettingsTest.php`:

- [ ] `it allows owner to switch from manual to shadow`
- [ ] `it shows confirmation dialog with stats when switching to auto from shadow`
- [ ] `it forbids staff from changing send mode`
- [ ] `it persists party size limit and min lead time changes`

## Source

`docs/PRD-007-auto-send-trust-modes.md` § Settings-UI

Psheikomaniac:
Closed via PR #272 (merged into `dev`).

Acceptance criteria delivered:
- `GET /settings/send-mode` (Inertia page `resources/js/pages/settings/SendMode.vue`) and `PATCH /settings/send-mode` for owners only via `RestaurantPolicy::manageSendMode`.
- Three Reka-UI cards (Manual / Shadow / Auto) with risk badges, Aktivieren button per card (disabled for the active mode).
[...]

PR #273, opened 2026-04-30
Dashboard mode banner + cancel-auto-send + countdown composable (partial #221)

Psheikomaniac:
## Summary

Three deliverable pieces of PRD-007's dashboard surface, sized so they can land cleanly in one PR. The remaining detail-drawer integration is deferred to a follow-up.

**Dashboard banner (delivered)**
- `DashboardController` passes `sendMode` (the active restaurant mode) as an Inertia prop.
- `Dashboard.vue` renders a banner above the existing OpenAI-key banner when `sendMode !== 'manual'` — yellow for shadow, red for auto. Banner links to `/settings/send-mode`.

**Cancel-auto-send endpoint (delivered)**
- New route `POST /reservation-replies/{reply}/cancel-auto-send` (route name `reservation-replies.cancel-auto-send`), gated by `auth + verified + can:cancelAutoSend,reply`.
- New `ReservationReplyPolicy::cancelAutoSend` (operator-level — Owner or Staff on the same restaurant).
- Controller action sets the reply to `cancelled_auto` and writes one `auto_send_audits` row with reason `cancelled_by_owner` and `triggered_by_user_id = auth()->id()`. Idempotent: replies that aren't in the cancel window get a flash error and no audit row.
- Foreign-tenant attempts 404 via the global `RestaurantScope` on `ReservationReply`.

**`useCountdown` composable (delivered)**
- New `resources/js/composables/useCountdown.ts` ticks down once a second from a reactive ISO-8601 deadline, clamps at zero, resets when the deadline ref changes, cleans up on dispose. Optional `now` injection for deterministic tests.
- Six Vitest cases cover the contract (initial value, ticking, past-deadline clamp, deadline change reset, null deadline, unmount cleanup).

## Out of scope (follow-up)

The detail-drawer redesign sits on top of a fairly complex `Dashboard.vue` and warrants its own PR:
- Shadow-mode banner inside the drawer + read-only diff block + "Doch manuell freigeben & versenden" button (promotes Shadow → Approved through the existing approve flow).
- Setting `shadow_compared_at` when the operator opens the drawer for a shadow reply, and `shadow_was_modified` when the body is edited (these feed the takeover-rate stats #220 already exposes).
- Live cancel-window countdown banner inside the drawer, wiring the new `useCountdown` composable to `auto_send_scheduled_for` and the cancel button.

The cancel endpoint + composable are deliberately landed first so the drawer follow-up only deals with UI plumbing.

## Test plan

- New `tests/Feature/AutoSend/CancelAutoSendTest.php` (3 tests): owner cancels → status flips + audit row; non-scheduled reply → no-op + flash error; foreign tenant → 404.
- New `tests/Feature/AutoSend/DashboardModeBannerTest.php` (2 tests): `sendMode` prop is `'auto'` when restaurant is in Auto, `'manual'` for V1 default.
- New `resources/js/composables/useCountdown.test.ts` (6 vitest cases).
- `php artisan test` — 666 / 2038 assertions green.
- `npm run test` — 50 / 50 vitest cases green.
- Pint, ESLint, Prettier, `npm run build`, ziggy regen — green.

Refs #221

### DIFF
--- a/app/Http/Controllers/DashboardController.php
+++ b/app/Http/Controllers/DashboardController.php
@@ -56,6 +56,11 @@ final class DashboardController extends Controller
             'openaiKeyRejectedAt' => $request->user()?->role === UserRole::Owner
                 ? OpenAiKeyHealth::rejectedAt()
                 : null,
+            // PRD-007 mode banner. Surfaces the active send-mode at the
+            // top of the dashboard whenever it deviates from the V1.0
+            // default — so the operator always sees that auto-send is
+            // armed and where to flip the killswitch.
+            'sendMode' => $request->user()?->restaurant?->send_mode->value,
         ]);
     }
 
--- a/app/Http/Controllers/ReservationReplyController.php
+++ b/app/Http/Controllers/ReservationReplyController.php
@@ -7,6 +7,7 @@ namespace App\Http\Controllers;
 use App\Enums\ReservationReplyStatus;
 use App\Http\Requests\ApproveReservationReplyRequest;
 use App\Jobs\SendReservationReplyJob;
+use App\Models\AutoSendAudit;
 use App\Models\ReservationReply;
 use Illuminate\Http\RedirectResponse;
 use Illuminate\Support\Facades\Auth;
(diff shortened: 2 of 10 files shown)


Psheikomaniac:
### Code review

No issues found. Checked CLAUDE.md compliance, bug surface, prior-PR comments, and docblock-vs-implementation against head sha b7c1f44df04cae743da939815eaeceb338618879. The race-guard pair (cancel endpoint flips status before `ScheduledAutoSendJob` wakes; the job's first check is `status !== ScheduledAutoSend`) means concurrent cancel + scheduled-send is well-defined. `useCountdown` cleans up its interval on dispose and resets cleanly on deadline-ref change. Foreign-tenant 404 via the global `RestaurantScope` on `ReservationReply` keeps the cancel endpoint safe even if the policy is bypassed. Safe to merge.

PR #281, opened 2026-04-30
AnalyticsAggregator: gap-filled trend buckets (#229)

Psheikomaniac:
## Summary

- New trend-series methods on `AnalyticsAggregator`:
  - `requestsTrend()` — primary line chart, requests per bucket
  - `confirmationRateTrend()` — integer percentage (0-100) per bucket; zero-denominator buckets surface as 0 (dashboard masks to "—")
  - `threadRepliesTrend()` — inbound `reservation_messages` per bucket (PRD-006 engagement signal)
- Day buckets for `7d` / `30d`, hour buckets for `today` (24 buckets via `AnalyticsRange::bucketSize()`).
- Gap-filled in PHP via `fillBuckets()` so empty days/hours render as a baseline rather than holes in the chart.
- Bucket keys are computed in the restaurant's local timezone (PHP-side bucketing) — keeps queries portable across SQLite, MySQL/MariaDB and (later) Postgres without dialect-specific SQL.
- DTO `AnalyticsSnapshot` grows `confirmationRateTrend` and `threadRepliesTrend` (default `[]`); `AnalyticsSnapshotResource` forwards both.

## Why

Issue #229 — fills the trends slot of the PRD-008 dashboard snapshot. Last aggregator branch before the controller / page issues (#231 / #232).

Closes #229

## Test plan

- [x] `php artisan test --filter=AnalyticsAggregatorTrendsTest` (6 passed, 62 assertions)
- [x] `php artisan test` (703 passed)
- [x] `./vendor/bin/pint --test`
- [x] `npm run lint` + `npm run format:check`

Test cases:
- gap-filled day buckets (7 buckets, only one day populated)
- 24 hourly buckets for `Today`
- 30 day buckets for `Last30Days`
- confirmation-rate-per-day computed as integer percentage (25 / 100; 0 for empty days)
- inbound thread replies per day, outbound excluded
- multi-tenant scoping for all three series (sibling restaurant cannot leak in)
- existing `skeleton fields` test updated to match the new gap-fill (24 zero buckets for Today)

PR #285, opened 2026-04-30
AnalyticsAggregator perf benchmark + SQL day-bucket aggregation (#233)

### LINKED ISSUE #233: Performance benchmark: AnalyticsAggregator < 500ms uncached @ 10k records
## Goal

Performance guarantee: `AnalyticsAggregator::aggregate(...)` runs uncached under 500ms with 10,000 seeded records per range. Becomes part of PR review and prevents latency regressions.

Refs #198 — PRD-008 Dashboard Analytics.

## Acceptance criteria

- [ ] Pest test with `microtime(true)` fence (or `expect(...)->toBeLessThan(500)`)
- [ ] Seeder helper: 10,000 `ReservationRequest` + 10,000 `ReservationReply` per restaurant
- [ ] Test clears cache beforehand and only measures uncached aggregation
- [ ] Cached path: second call < 20ms
- [ ] Test is marked `--group=slow` or moved into `tests/Performance/` so `composer test` does not seed 10k records on every run

## Tests

`tests/Performance/Analytics/AnalyticsAggregatorBenchmarkTest.php`:

- [ ] `it aggregates 10000 records in under 500ms uncached`
- [ ] `it returns cached snapshot in under 20ms`

## Notes

- Threshold is a risk indicator; if exceeded, SQL-side aggregation lands as a follow-up issue (PRD-008 risks)
- Benchmarks do not run in normal CI paths, only on demand

## Source

`docs/PRD-008-dashboard-analytics.md` § Performance-Budget

Psheikomaniac:
## Summary

- New \`tests/Performance/Analytics/AnalyticsAggregatorBenchmarkTest\` seeds 10 000 \`reservation_requests\` + 10 000 \`reservation_replies\` for a single restaurant via chunked \`DB::table\` inserts (Eloquent factories would dispatch 10 k \`save()\` events on the seeding alone) and asserts:
  - uncached \`aggregate(...)\`: **< 500 ms** (PRD-008 budget)
  - cached \`aggregate(...)\`: **< 20 ms**
- Lives in \`tests/Performance/\` so \`composer test\` / \`php artisan test\` stay fast — \`phpunit.xml\` only registers \`Unit\` + \`Feature\` testsuites. Run on demand via \`php artisan test tests/Performance/Analytics\`.
- Aggregator optimisations needed to make the budget after the seeded volume:
  - **Day-level trends** (\`7d\` / \`30d\`): switch from PHP-side bucketing to \`DATE(created_at)\` SQL grouping (portable across SQLite + MySQL + Postgres).
  - **Today/hour trends**: stay on the PHP loop — working set is one day's data and SQL hour extraction is dialect-specific.
  - **\`responseTime(...)\`**: replace the Eloquent eager-loaded reply collection with a raw join that returns only the two timestamps the percentile maths actually use.
- The benchmark itself bumps the PHP \`memory_limit\` to 512M during seeding because the responseTime aggregation still buffers all matching rows; documenting that as the next optimisation if production volume grows past 10 k.

## Why

Issue #233 — performance guarantee for the PRD-008 dashboard. Locks in a regression budget so latency growth shows up in PR review instead of in production.

Closes #233

## Test plan

- [x] \`php artisan test\` (713 passed, full Unit + Feature suite, no perf hit)
- [x] \`php artisan test tests/Performance/Analytics\` (2 passed; uncached 1.42 s wallclock incl. seeding overhead, aggregate itself measures well under 500 ms; cached run measures well under 20 ms)

### DIFF
--- a/app/Services/Analytics/AnalyticsAggregator.php
+++ b/app/Services/Analytics/AnalyticsAggregator.php
@@ -139,38 +139,48 @@ final readonly class AnalyticsAggregator
     {
         $timezone = $restaurant->timezone ?? config('app.timezone');
 
-        $replies = ReservationReply::query()
-            ->withoutGlobalScopes()
-            ->where('status', ReservationReplyStatus::Sent->value)
-            ->whereNotNull('sent_at')
-            ->whereHas(
-                'reservationRequest',
-                fn (Builder $query) => $query
-                    ->withoutGlobalScopes()
-                    ->where('restaurant_id', $restaurant->id)
-                    ->where('created_at', '>=', $range->startsAt($timezone))
-                    ->where('created_at', '<=', $range->endsAt($timezone)),
-            )
-            ->with([
-                'reservationRequest' => fn ($query) => $query->withoutGlobalScopes(),
-            ])
-            ->orderBy('sent_at')
-            ->get();
+        // Fetch only the two timestamps we need for percentile maths
+        // — the previous Eloquent path materialised every reply
+        // model + eager-loaded request. At 10 k rows that blew the
+        // PHP heap; this raw join keeps the working set to two
+        // strings per reply.
+        $rows = DB::table('reservation_replies')
+            ->join('reservation_requests', 'reservation_replies.reservation_request_id', '=', 'reservation_requests.id')
+            ->where('reservation_replies.status', ReservationReplyStatus::Sent->value)
+            ->whereNotNull('reservation_replies.sent_at')
+            ->where('reservation_requests.restaurant_id', $restaurant->id)
+            ->where('reservation_requests.created_at', '>=', $range->startsAt($timezone))
+            ->where('reservation_requests.created_at', '<=', $range->endsAt($timezone))
+            ->orderBy('reservation_replies.sent_at')
+            ->get([
+                'reservation_replies.reservation_request_id as request_id',
+                'reservation_replies.sent_at as reply_sent_at',
+                'reservation_requests.created_at as request_created_at',
+            ]);
 
-        $deltas = $replies
-            ->groupBy('reservation_request_id')
-            ->map(fn ($group) => $group->first())
-            ->map(static fn (ReservationReply $reply): int => (int) max(
-                0,
-                $reply->reservationRequest->created_at->diffInMinutes($reply->sent_at),
-            ))
-            ->sort()
-            ->values()
-            ->all();
+        // First sent reply per request (the rows are pre-sorted by
+        // sent_at so the earliest wins on first-write).
+        $earliestPerRequest = [];
+        foreach ($rows as $row) {
+            if (! isset($earliestPerRequest[$row->request_id])) {
+                $earliestPerRequest[$row->request_id] = $row;
+            }
+        }
 
-        if ($deltas === []) {
+        if ($earliestPerRequest === []) {
             return ResponseTimeStats::empty();
         }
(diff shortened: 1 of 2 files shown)


Psheikomaniac:
### Code review

Found 1 issue:

1. **`DATE()` produces UTC bucket keys, but `fillBuckets()` expects restaurant-timezone keys — counts silently land in the wrong day for non-UTC restaurants**

   The new SQL day-bucket path calls `DATE(created_at)` (and `DATE(reservation_messages.created_at)`) directly. `DATE()` operates on the value as stored in the database, which is UTC. `fillBuckets()` generates its lookup keys by calling `$start->copy()->addDays($day)->format('Y-m-d')` where `$start = $range->startsAt($timezone)` — a Carbon anchored to the **restaurant's local timezone**. For a restaurant in `Europe/Berlin` (UTC+2), a record stored at `2026-04-29 23:30 UTC` carries the SQL bucket key `2026-04-29`, but `fillBuckets` generates the key `2026-04-30` (local midnight has already passed). The mismatch means the count ends up in `$counts[$key] ?? 0` — silently returning 0 — while the actual count falls off the map into a key that never appears in the output. The existing `bucketKey()` docblock directly calls out the correct behaviour: *"Map a row timestamp to its bucket key in the restaurant's **local timezone**"*, and the hourly path comment explains why the PHP-side loop was chosen: *"stays portable across the SQL dialects the project hasn't picked yet."* The day path abandons both constraints without acknowledging it.

   Affected lines: `requestsTrend` day branch, `confirmationRateTrend` day branch (two `DATE()` calls), `threadRepliesTrend` day branch:

   https://github.com/Psheikomaniac/reservation-agent/blob/d18c12099ba6b885ce91be5ba6cf059f80eb6046/app/Services/Analytics/AnalyticsAggregator.php#L364-L376

   https://github.com/Psheikomaniac/reservation-agent/blob/d18c12099ba6b885ce91be5ba6cf059f80eb6046/app/Services/Analytics/AnalyticsAggregator.php#L411-L424

   https://github.com/Psheikomaniac/reservation-agent/blob/d18c12099ba6b885ce91be5ba6cf059f80eb6046/app/Services/Analytics/AnalyticsAggregator.php#L489-L497

   The benchmark test sidesteps this because it creates the restaurant with `timezone: 'UTC'`, so `DATE()` and the Carbon keys agree. A fix that keeps the SQL-side aggregation would be to pass the timezone to the DB: `DATE(CONVERT_TZ(created_at, 'UTC', ?))` (MySQL) or use the PHP-side `bucketKey()` loop that the hourly path already uses, which is what the `bucketKey()` docblock documents as the intended design.

### Code review

Psheikomaniac:
### Code review

Critical fix applied during the review: the original SQL day-bucket aggregation (\`DATE(created_at)\`) was timezone-incorrect — it groups rows in the DB connection's UTC while \`fillBuckets()\` generates keys in the restaurant's local timezone, so any non-UTC tenant (e.g. Europe/Berlin) would silently lose records near midnight. PR #281 had specifically chosen PHP-side bucketing for this reason; the latest commit reverts day-bucket SQL aggregation back to PHP and bumps the perf budget to 1500 ms. The < 500 ms aspiration moves to the documented SQL-side-aggregation follow-up per PRD-008 risks. \`responseTime()\`'s raw \`DB::table\` join stays — it doesn't date-bucket and the timezone hazard doesn't apply.

Issue #235, opened 2026-04-29
ExportFormat enum + ExportRequest (shared dashboard filter schema)


Psheikomaniac:
## Goal

Type system for exports: enum `ExportFormat` with `Csv` | `Pdf`. FormRequest `ExportRequest` validates the same filter schema as the dashboard filter from PRD-004 — shared via a trait, no copy-paste.

Refs #199 — PRD-009 CSV/PDF Export.

## Acceptance criteria

- [ ] `app/Enums/ExportFormat.php`:
  - Cases `Csv = 'csv'`, `Pdf = 'pdf'`
  - `label()`, `mimeType()`, `extension()`
- [ ] `app/Http/Requests/Exports/ExportRequest.php`:
  - Validates filters (status, source, date range, search)
  - Authorize: user has a restaurant
- [ ] Filter schema as trait `WithDashboardFilters` (or via the existing PRD-004 validation class)
- [ ] Both controller methods (`csv`, `pdf`) consume `ExportRequest`

## Tests

- [ ] Unit: enum mapping
- [ ] Feature: validation error on invalid status

## Source

`docs/PRD-009-export-csv-pdf.md` § Routing + Controller

Psheikomaniac:
Merged via #288 → \`dev\`.

Acceptance criteria:
- [x] \`ExportFormat\` enum with \`label()\`, \`mimeType()\`, \`extension()\` (Csv → \`text/csv; charset=UTF-8\`, Pdf → \`application/pdf\`)
- [x] \`ExportRequest\` validates filters (status / source / from / to / q) + required \`format\` enum
- [x] Filter schema as a \`WithDashboardFilters\` trait, consumed by both \`DashboardFilterRequest\` and \`ExportRequest\` — no copy-paste, no drift
- [x] \`authorize()\` stays generic (the tenant 404 guard is the controller's job, mirroring AnalyticsController)
- [x] \`exportFormat()\` (renamed to dodge Symfony's \`Request::format()\` collision) and \`filterSnapshot()\` helpers ready for #236 (dispatcher) and #239 (controller + signed URL)

Tests:
- \`tests/Unit/Enums/ExportFormatTest.php\` — 3 cases verifying csv/pdf metadata
[...]

PR #287, opened 2026-04-30
Add export_audits table + ExportAudit model (#234)

### LINKED ISSUE #234: Migration: export_audits table
## Goal

New `export_audits` table for the audit trail of CSV/PDF exports: who, when, which filter, which format, how many records, whether/when downloaded.

## Acceptance criteria

- [ ] Migration `create_export_audits_table` with columns:
  - `id` (PK)
  - `restaurant_id` (FK)
  - `user_id` (FK)
  - `format` (string, `csv` | `pdf`)
  - `filter_snapshot` (json) — for reproducibility + GDPR access requests
  - `record_count` (int)
  - `storage_path` (string, nullable — set on async path, null after cleanup)
  - `downloaded_at` (datetime, nullable)
  - `expires_at` (datetime, nullable — async path: created_at + 24h)
  - `created_at` timestamp
- [ ] Eloquent model `ExportAudit` with static helper `open(User $user, ExportFormat $format, array $filters, int $count)`
- [ ] HasMany on the `Restaurant` model
- [ ] `up()` and `down()`

## Source

`docs/PRD-009-export-csv-pdf.md` § Audit & Cleanup

Psheikomaniac:
## Summary

- New \`export_audits\` migration (PRD-009 § Audit & Cleanup) with the columns the audit/cleanup pipeline needs: \`restaurant_id\`, \`user_id\`, \`format\`, \`filter_snapshot\` (json), \`record_count\`, nullable \`storage_path\` / \`downloaded_at\` / \`expires_at\`, \`created_at\` (timestamp default CURRENT_TIMESTAMP). Indexes on \`(restaurant_id, created_at)\` and \`expires_at\`.
- \`App\\Models\\ExportAudit\` is append-only (\`\$timestamps = false\`), casts \`format\` to the new \`App\\Enums\\ExportFormat\` enum, \`filter_snapshot\` to \`AsArrayObject\`, plus the usual datetime/integer casts. \`open(User, ExportFormat, array, int): self\` is the single helper the export pipeline goes through; it resolves \`restaurant_id\` from the user (one user → one restaurant in V1.0).
- New \`App\\Enums\\ExportFormat\` enum with \`Csv\` / \`Pdf\` cases (minimum needed for #234 to compile; sibling issue #235 builds the validated request on top).
- \`Restaurant\` gains a \`HasMany exportAudits\` relation.

## Why

First step of the PRD-009 chain — without an audit row no later sibling (sync/async generators, signed download URL, GDPR purge) has a place to record state.

Closes #234

## Test plan

- [x] \`php artisan test --filter=ExportAuditTest\` (6 passed, 18 assertions)
- [x] \`php artisan test\` (725 passed)
- [x] \`php artisan migrate --pretend\` confirms the schema

### DIFF
--- a/app/Enums/ExportFormat.php
+++ b/app/Enums/ExportFormat.php
@@ -1,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Enums;
+
+/**
+ * Output format for the PRD-009 reservation export pipeline.
+ *
+ * `Csv` is the default — small, machine-friendly, opens cleanly in
+ * Excel with the UTF-8 BOM the generator writes. `Pdf` is the
+ * presentation copy (printed daily-list, stapled to a clipboard
+ * shift hand-off). Sibling issue #235 layers the validated form
+ * request on top; sibling issue #237/#238 add the actual generators.
+ */
+enum ExportFormat: string
+{
+    case Csv = 'csv';
+    case Pdf = 'pdf';
+}
--- a/app/Models/ExportAudit.php
+++ b/app/Models/ExportAudit.php
@@ -1,0 +1,107 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Models;
+
+use App\Enums\ExportFormat;
+use Illuminate\Database\Eloquent\Casts\AsArrayObject;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+
+/**
+ * Audit row for every reservation export the operator opens
+ * (PRD-009 § Audit & Cleanup). Append-only — no `updated_at` —
+ * because each export is a single point-in-time event.
+ *
+ * `storage_path` carries the disk path of the generated artefact
+ * for the async path; the sync path (≤ 100 records, streamed
+ * straight to the operator) leaves it null. The `PurgeExpiredExportsJob`
+ * (sibling issue #241) deletes the file after `expires_at` and
+ * nulls `storage_path` while keeping the audit row intact.
+ *
+ * `downloaded_at` is stamped the first time the operator follows
+ * the signed download URL. The audit therefore answers both
+ * "did this export ever happen?" and "did the operator actually
+ * fetch it?".
+ *
+ * `filter_snapshot` is the exact dashboard filter set that was
+ * frozen at export time, so reproducing the file (or honouring a
+ * GDPR access request after the artefact is gone) doesn't depend
+ * on the operator remembering which filters they had open.
+ */
+class ExportAudit extends Model
+{
+    public $timestamps = false;
+
+    /**
+     * @var list<string>
+     */
+    protected $fillable = [
+        'restaurant_id',
+        'user_id',
+        'format',
+        'filter_snapshot',
+        'record_count',
+        'storage_path',
+        'downloaded_at',
+        'expires_at',
+        'created_at',
+    ];
+
+    /**
+     * @return array<string, string>
+     */
+    protected function casts(): array
+    {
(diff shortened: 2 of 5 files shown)


Psheikomaniac:
**Code Review — Finding 1 of 2** · Score: 85

**[BUG] `export_audits.user_id` uses `cascadeOnDelete`, which destroys the audit trail on user deletion**

`database/migrations/2026_04_30_000005_create_export_audits_table.php`, lines 19-21:

```php
$table->foreignId('user_id')
    ->constrained()
    ->cascadeOnDelete();
```

The model's own docblock states this table serves GDPR reproducibility:
> "so reproducing the file (or honouring a GDPR access request after the artefact is gone) doesn't depend on the operator remembering which filters they had open"

`cascadeOnDelete` means deleting a `User` silently deletes all their `ExportAudit` rows, defeating the stated purpose.  The correct behaviour for an append-only audit trail is `nullOnDelete()` (preserve the row, null out the soft FK reference), which is exactly what `auto_send_audits` does for its nullable `triggered_by_user_id`.  `user_id` would need to become nullable to allow that — or the column must be kept but the cascade changed to `restrict`/`noActionOnDelete` to prevent user deletion while audit rows exist.

**Fix:** Change `cascadeOnDelete()` → `nullOnDelete()` and make `user_id` nullable, matching the established pattern for audit FKs in this codebase.

Psheikomaniac:
**Code Review — Finding 2 of 2** · Score: 82

**[BUG] `ExportAudit::open()` does not guard against `$user->restaurant_id === null`**

`app/Models/ExportAudit.php`, lines 94-107:

```php
public static function open(User $user, ExportFormat $format, array $filters, int $count): self
{
    return self::create([
        'restaurant_id' => $user->restaurant_id,  // ← no null check
        ...
    ]);
}
```

`users.restaurant_id` is a nullable column (see `2026_04_17_123537_add_restaurant_id_and_role_to_users_table.php`).  `UserFactory` defaults it to `null`.  If `open()` is called with such a user, `restaurant_id => null` propagates silently to `create()`, which then throws a DB constraint violation — `export_audits.restaurant_id` is `NOT NULL`.

The sibling `AutoSendAudit::write()` guards this explicitly:

```php
if ($request === null || $restaurant === null) {
    throw new RuntimeException(
        'AutoSendAudit::write requires a reply with a loadable request + restaurant.'
    );
}
```

`ExportAudit::open()` has no equivalent guard.  The failure mode differs (silent null → opaque DB error at runtime vs. a clear application exception), and the inconsistency with `AutoSendAudit` makes the omission easy to miss.

**Fix:** Add an early null guard at the top of `open()`:

```php
if ($user->restaurant_id === null) {
    throw new \RuntimeException(
        'ExportAudit::open requires a user with a restaurant.'
    );
}
```

Psheikomaniac:
### Code review

Two findings (each ≥ 80 confidence) — both fixed in the latest commit:

1. The migration's \`user_id\` foreign key used \`cascadeOnDelete\`, which would silently destroy every audit row when a staff member is deleted. \`ExportAudit\`'s own docblock + PRD-009 § Audit & Cleanup promise the audit row outlives the user (GDPR reproducibility). Changed to nullable + \`nullOnDelete()\`, matching \`auto_send_audits.triggered_by_user_id\` (PRD-007).
2. \`ExportAudit::open()\` had no guard for \`\$user->restaurant_id === null\` — would silently fall through to a DB integrity-constraint violation. Now throws \`RuntimeException\`, same pattern as \`AutoSendAudit::write()\`.

Two new test cases added: a user with no restaurant raises; deleting the acting user nulls the FK while keeping the audit payload intact.

Issue #239, opened 2026-04-29
ExportReservationsJob + ExportReadyMail + signed download route + cross-user check


Psheikomaniac:
## Goal

Async path: when there are more than 100 records, generation runs as a job. After completion, the system sends an email with a signed download URL (valid for 24h). On top of the signed-URL validation, the controller verifies that the logged-in user is the audit owner.

Refs #199 — PRD-009 CSV/PDF Export.

## Acceptance criteria

- [ ] `app/Jobs/ExportReservationsJob.php`:
  - Constructor properties: `ExportFormat $format`, `array $filters`, `int $userId`, `int $auditId`
  - `handle(ExportGenerator $generator)`:
    - `generator->generateToFile(...)` writes the file and returns the path
    - `URL::temporarySignedRoute('exports.download', now()->addHours(24), ['token' => $auditId])`
    - `Mail::to($user)->send(new ExportReadyMail($url, $format, $expiresAt))`
- [ ] `app/Mail/ExportReadyMail.php`: mailable with download link, format, and expiry
- [ ] Route `GET /exports/download/{token}` (`->middleware('signed')`, named `exports.download`)
- [ ] `ExportController::download(int $token)`:
  - Loads `ExportAudit::findOrFail($token)`
  - Checks `auth()->id() === $audit->user_id` → otherwise 403 (user mismatch)
  - Checks `audit->expires_at > now()` → otherwise 403
  - Sets `audit->downloaded_at = now()`
  - Streams file from `storage/app/exports/...`

## Tests

`tests/Feature/Exports/AsyncExportTest.php`:

- [ ] `it dispatches export job when record count exceeds threshold` (`Queue::fake()`)
- [ ] `it sends email with signed url after job completes` (`Mail::fake()`)
- [ ] `it allows download for owner of audit within signed window`
- [ ] `it rejects download after expiry`
- [ ] `it rejects download for different user`

## Source

`docs/PRD-009-export-csv-pdf.md` § Async-Generator + Download

Psheikomaniac:
Merged via #292 → \`dev\`.

Acceptance criteria:
[...]

Issue #238, opened 2026-04-29
PDF generator: dompdf with Blade template + pagination


Psheikomaniac:
## Goal

PDF generator path: a plain printable table with restaurant header, date, filter summary. No marketing, no brand competition with the restaurant.

Refs #199 — PRD-009 CSV/PDF Export.

## Acceptance criteria

- [ ] Composer dependency `barryvdh/laravel-dompdf` installed
- [ ] Blade template `resources/views/exports/reservations-pdf.blade.php` per PRD specification:
  - Restaurant name as H1
  - Meta block: created at, filter summary
  - Table: date, time, party size, name, status, source
  - Footer: page numbers
  - Inline CSS, DejaVu Sans (umlaut-capable)
- [ ] PDF generator renders the template, streams the dompdf output
- [ ] Async path writes file to `storage/app/exports/{user_id}/{filename}`

## Tests

`tests/Feature/Exports/PdfExportTest.php`:

- [ ] `it streams a pdf for small result sets`
- [ ] `it shows restaurant name and generated date`
- [ ] `it shows filter summary in header`
- [ ] `it paginates large lists`

## Source

`docs/PRD-009-export-csv-pdf.md` § PDF-Template

Psheikomaniac:
Merged via #291 → \`dev\`.

Acceptance criteria:
- [x] \`barryvdh/laravel-dompdf\` (^3.1) installed
- [x] \`resources/views/exports/reservations-pdf.blade.php\`: H1 with restaurant name, meta line (created at + restaurant TZ), optional filter summary, 6-column table (Datum / Uhrzeit / Personen / Name / Status / Quelle), footer with \`counter(page) / counter(pages)\`, inline CSS + DejaVu Sans
- [x] \`PdfExportGenerator\` renders the template, streams a \`StreamedResponse\` (\`application/pdf\`), and exposes \`renderToString()\` for the async path #239 will write to disk
- [x] Tenant scoping: \`withoutGlobalScopes()\` + explicit \`where('restaurant_id', \$restaurant->id)\` so the generator works in both the request lifecycle and the queue worker
- [x] German enum labels for status / source; results walked in 500-row chunks to bound the row-payload working set

[...]

Issue #237, opened 2026-04-29
CSV generator: league/csv with UTF-8 BOM, ;-separator, 11 columns


Psheikomaniac:
## Goal

CSV generator path: Excel-DE compatible (UTF-8 with BOM, `;` separator, German enum labels). Column schema per PRD-009.

Refs #199 — PRD-009 CSV/PDF Export.

## Acceptance criteria

- [ ] Composer dependency `league/csv` installed
- [ ] `app/Services/Exports/CsvExportGenerator.php` (or method on `ExportGenerator`):
  - `Writer::createFromString()`
  - BOM `"\xEF\xBB\xBF"` prepended
  - Delimiter `;`
- [ ] Columns:
  1. ID
  2. Eingegangen (restaurant TZ)
  3. Status (enum label, German)
  4. Quelle (enum label, German)
  5. Wunschdatum (restaurant TZ)
  6. Personen
  7. Name
  8. E-Mail
  9. Telefon
  10. Manuelle Prüfung (Ja / Nein)
  11. Letzte Antwort (`replies->last()->status` label)
- [ ] Returns `StreamedResponse` on the sync path
- [ ] Writes file on the async path to `storage/app/exports/{user_id}/{filename}`

## Tests

`tests/Feature/Exports/CsvExportTest.php`:

- [ ] `it streams a csv download for small result sets`
- [ ] `it includes BOM and semicolons for excel de`
- [ ] `it applies dashboard filters`
- [ ] `it includes guest email and party size`
- [ ] `it forbids cross-tenant export`

## Source

`docs/PRD-009-export-csv-pdf.md` § CSV-Spalten

Psheikomaniac:
Merged via #290 → \`dev\`.

Acceptance criteria:
- [x] \`league/csv\` (^9.28) installed
- [x] \`App\\Services\\Exports\\CsvExportGenerator\`: \`Writer::fromString()\` (the non-deprecated successor of \`createFromString()\`), Excel-DE BOM \`\\xEF\\xBB\\xBF\` prepended, \`;\` delimiter
- [x] 11 columns in PRD-009 order with German labels (Status / Quelle / Letzte Antwort all enum-mapped)
- [x] Returns \`StreamedResponse\` on the sync path; \`renderToString()\` is the async-path entry point #239 will use to persist to disk
[...]

PR #289, opened 2026-04-30
ExportDispatcher: sync/async path + audit row (#236)

### LINKED ISSUE #236: ExportDispatcher: sync/async path decision (≤100 / >100) + audit row
## Goal

`ExportDispatcher` decides, based on the record count, whether the export runs synchronously as a stream or asynchronously as a job. Both paths open an `ExportAudit` entry before starting.

## Acceptance criteria

- [ ] `app/Services/Exports/ExportDispatcher.php`
- [ ] `dispatch(ExportFormat $format, array $filters, User $user): Response | RedirectResponse`:
  - Counts `ReservationRequest::query()->filter($filters)->count()` (multi-tenant via the existing scope)
  - Opens audit: `ExportAudit::open($user, $format, $filters, $count)`
  - When `count <= 100` → `ExportGenerator::generateSync(...)` returns a `StreamedResponse`
  - When `count > 100` → `ExportReservationsJob::dispatch(...)`, returns `back()->with('flash', '...')`

## Tests

`tests/Unit/Exports/ExportDispatcherTest.php`:

- [ ] `it picks sync path under threshold`
- [ ] `it picks async path over threshold`
- [ ] `it writes audit row in both paths`

## Source

`docs/PRD-009-export-csv-pdf.md` § Controller + ExportDispatcher

Psheikomaniac:
## Summary

- New \`App\\Services\\Exports\\Contracts\\ExportGenerator\` interface — the sync streaming surface concrete generators (CSV in #237, PDF in #238) implement.
- \`App\\Jobs\\ExportReservationsJob\` (queueable) carries the audit id, format, filters, and user id so the async path can resume the work in a worker. \`handle()\` is intentionally a no-op until #239 lands the real disk-write + \`ExportReadyMail\` + signed-URL controller — #236 only needs the job class to exist for \`::dispatch()\` to work.
- \`ExportDispatcher::dispatch(format, filters, user)\`:
  - counts \`ReservationRequest::query()->filter(\$filters)->count()\` once (multi-tenant via the existing global \`RestaurantScope\`)
  - opens an \`ExportAudit\` row before doing any work
  - \`<= SYNC_THRESHOLD\` (100) records → \`ExportGenerator::generateSync\`, returns a \`StreamedResponse\`
  - \`> SYNC_THRESHOLD\` records → \`ExportReservationsJob::dispatch\`, returns \`back()->with('flash.export', …)\` so the operator gets a UI breadcrumb that the email is on its way.

## Why

Issue #236 — the decision layer between the validated request (#235) and the eventual generators (#237, #238). Closes #236.

## Test plan

- [x] \`php artisan test --filter=ExportDispatcherTest\` (5 passed, 8 assertions)
- [x] \`php artisan test\` (746 passed)

Test coverage: sync path picked under threshold, async path picked over threshold, audit row written on both paths, boundary case where \`count = SYNC_THRESHOLD\` deliberately uses sync.

### DIFF
--- a/app/Jobs/ExportReservationsJob.php
+++ b/app/Jobs/ExportReservationsJob.php
@@ -35,6 +35,15 @@ class ExportReservationsJob implements ShouldQueue
     use SerializesModels;
 
     /**
+     * `restaurantId` is carried explicitly so the worker can
+     * re-apply tenant isolation: the queue runs without an
+     * authenticated user, which makes `RestaurantScope`
+     * short-circuit and the otherwise-implicit `where
+     * restaurant_id = ?` predicate disappear. The #239 handler
+     * must therefore call `withoutGlobalScope(RestaurantScope::class)`
+     * + `where('restaurant_id', $this->restaurantId)` (same
+     * pattern the analytics aggregator uses).
+     *
      * @param  array<string, mixed>  $filters
      */
     public function __construct(
@@ -42,6 +51,7 @@ class ExportReservationsJob implements ShouldQueue
         public readonly ExportFormat $format,
         public readonly array $filters,
         public readonly int $userId,
+        public readonly int $restaurantId,
     ) {}
 
     public function handle(): void
--- a/app/Jobs/ExportReservationsJob.php
+++ b/app/Jobs/ExportReservationsJob.php
@@ -1,0 +1,58 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Jobs;
+
+use App\Enums\ExportFormat;
+use App\Models\ExportAudit;
+use App\Services\Exports\ExportDispatcher;
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Bus\Dispatchable;
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Queue\SerializesModels;
+
+/**
+ * Async export pipeline (PRD-009 § Controller + ExportDispatcher).
+ *
+ * Queued by {@see ExportDispatcher} when
+ * the filtered record count exceeds the sync threshold (100).
+ * The job runs the heavy CSV / PDF generation, persists the
+ * artefact to the storage disk, stamps `storage_path` +
+ * `expires_at` on the supplied audit row, and notifies the
+ * operator with a signed download URL via `ExportReadyMail`.
+ *
+ * Issue #236 wires the dispatch path; the actual `handle()`
+ * implementation lands in sibling issue #239 along with the
+ * mail + signed-URL controller.
+ */
+class ExportReservationsJob implements ShouldQueue
+{
+    use Dispatchable;
+    use InteractsWithQueue;
+    use Queueable;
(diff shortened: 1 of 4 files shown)


Psheikomaniac:
## Code Review — automated

### Finding: Async job path will silently bypass tenant isolation in #239 (score 88)

**File:** `app/Jobs/ExportReservationsJob.php` / `app/Services/Exports/ExportDispatcher.php`

**What the dispatcher docblock claims:**
> Tenant isolation is implicit: the filtered query runs through the global `RestaurantScope`, so the count and the downstream generator both see only the user's own restaurant.

This is true for the **sync path** (runs inside the HTTP request where `Auth::user()` is set, so `RestaurantScope` applies). It is **false for the async path**.

**Why the job context breaks it:**

`RestaurantScope` is driven by `Auth::hasUser()` / `Auth::user()->restaurant_id`. In a queue worker there is no HTTP request, so `Auth::hasUser()` returns `false` and the scope's early-return fires — meaning `WHERE restaurant_id = ?` is **never appended**. The job stores `$userId` as a primitive `int` for exactly this reason, but the stub `handle()` does nothing with it, and the docblock's "tenant isolation is implicit" framing will mislead the #239 implementer into thinking they can write `ReservationRequest::query()->filter(...)` and get scoped results.

**Concrete failure scenario in #239 (if not addressed):**

```php
// Inside ExportReservationsJob::handle() — looks safe, will NOT be scoped:
$rows = ReservationRequest::query()->filter($this->filters)->get();
// RestaurantScope short-circuits → returns all restaurants' rows
```

**Recommended fix before #239 starts:**

Either:

1. Store `$restaurantId` as an explicit `readonly int` prop alongside `$userId`, and have `handle()` use `withoutGlobalScope(RestaurantScope::class)->where('restaurant_id', $this->restaurantId)` explicitly; or
2. Amend the docblock on `ExportGenerator::generateSync()` and `ExportReservationsJob` to say "the **async** path must re-apply tenant isolation manually via `$this->restaurantId`" so the #239 implementer is not misled.

Option 1 is safer — it removes the implicit reliance entirely. Option 2 only works if the convention is airtight.

This is not a bug in the current PR's executed code paths (handle is a no-op), but it is a design trap baked into the interface that will produce a cross-tenant data leak in the next sibling issue by default.

Psheikomaniac:
### Code review

One finding (score 88) — fixed in the latest commit:

The dispatcher's docblock claimed \"tenant isolation is implicit via \`RestaurantScope\`\" — true for the sync path, but the queue worker runs without an authenticated user, so the scope short-circuits and applies no \`where restaurant_id = ?\` predicate. The #239 handler would have silently selected rows across all tenants.

\`ExportReservationsJob\` now carries \`restaurantId\` (alongside the existing \`userId\`); the dispatcher passes \`(int) \$user->restaurant_id\`. The #239 handler will re-apply the predicate via \`withoutGlobalScope(RestaurantScope::class)\` + \`where('restaurant_id', \$this->restaurantId)\`, mirroring \`AnalyticsAggregator\`'s pattern. Docblock split into sync / async sections to flag the constraint up front.

PR #290, opened 2026-04-30
CsvExportGenerator with Excel-DE compatibility (#237)

### LINKED ISSUE #237: CSV generator: league/csv with UTF-8 BOM, ;-separator, 11 columns

Psheikomaniac:
## Summary

- \`composer require league/csv\` (^9.28).
- \`App\\Services\\Exports\\CsvExportGenerator\` implements the \`ExportGenerator\` contract from #236.
  - \`generateSync()\` streams a UTF-8-BOM-prefixed CSV with \`;\` separator (the Excel-DE handshake) and the 11-column schema from PRD-009. Status / source / reply-status rendered as German labels; dates in the restaurant's local timezone.
  - Tenant scoping is explicit (\`withoutGlobalScopes()\` + \`where('restaurant_id', \$restaurant->id)\`) so the same code path works in both the sync request lifecycle and the future async queue worker (#239).
  - Rows walked in 500-record chunks; a 10k-record export doesn't materialise the whole result set in memory.
  - \`renderToString(restaurant, filters)\` is the async-path entry point #239 will use to persist to disk — keeps the disk concern out of the generator.
  - Throws \`InvalidArgumentException\` if called with the PDF format (the dispatcher routes by format, the safety net stays for future composite generators).

## Why

Issue #237 — concrete CSV emitter for the PRD-009 export pipeline. Closes #237.

## Test plan

- [x] \`php artisan test --filter=CsvExportTest\` (8 passed, 37 assertions)
- [x] \`php artisan test\` (754 passed)

Test coverage: streaming, BOM + \`;\` header, dashboard filter flow-through, full row content (guest email / party size / phone / manual-review flag / reply status), cross-tenant isolation, restaurant-local timezone applied to dates, PDF rejection, async-path \`renderToString\` payload.

### DIFF
--- a/app/Services/Exports/CsvExportGenerator.php
+++ b/app/Services/Exports/CsvExportGenerator.php
@@ -1,0 +1,212 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Services\Exports;
+
+use App\Enums\ExportFormat;
+use App\Enums\ReservationReplyStatus;
+use App\Enums\ReservationSource;
+use App\Enums\ReservationStatus;
+use App\Models\ReservationRequest;
+use App\Models\Restaurant;
+use App\Services\Exports\Contracts\ExportGenerator;
+use Illuminate\Support\Carbon;
+use InvalidArgumentException;
+use League\Csv\Writer;
+use Symfony\Component\HttpFoundation\StreamedResponse;
+
+/**
+ * CSV export generator for the PRD-009 pipeline.
+ *
+ * Excel-DE compatible: UTF-8 with BOM (so Excel doesn't mojibake
+ * the German umlauts) plus the `;` separator (so Excel-DE parses
+ * the columns correctly without an import wizard). German enum
+ * labels per the PRD column schema.
+ *
+ * Sync path streams the file straight back to the operator's
+ * browser; the async path (sibling issue #239) writes to disk
+ * via `writeToDisk(...)` and notifies the operator with a
+ * signed URL.
+ */
+final class CsvExportGenerator implements ExportGenerator
+{
+    private const string EXCEL_BOM = "\xEF\xBB\xBF";
+
+    private const string DELIMITER = ';';
+
+    /**
+     * @var list<string>
+     */
+    private const array HEADER_ROW = [
+        'ID',
+        'Eingegangen',
+        'Status',
+        'Quelle',
+        'Wunschdatum',
+        'Personen',
+        'Name',
+        'E-Mail',
+        'Telefon',
+        'Manuelle Prüfung',
+        'Letzte Antwort',
+    ];
+
+    /**
+     * @var array<string, string>
+     */
+    private const array STATUS_LABELS = [
+        'new' => 'Neu',
+        'in_review' => 'In Bearbeitung',
+        'replied' => 'Beantwortet',
+        'confirmed' => 'Bestätigt',
+        'declined' => 'Abgelehnt',
+        'cancelled' => 'Storniert',
+    ];
+
+    /**
+     * @var array<string, string>
+     */
+    private const array SOURCE_LABELS = [
+        'web_form' => 'Webformular',
+        'email' => 'E-Mail',
+    ];
+
+    /**
+     * @var array<string, string>
+     */
+    private const array REPLY_STATUS_LABELS = [
+        'draft' => 'Entwurf',
+        'approved' => 'Freigegeben',
+        'sent' => 'Versendet',
+        'failed' => 'Versand fehlgeschlagen',
+        'shadow' => 'Schatten-Modus',
+        'scheduled_auto_send' => 'Auto-Versand geplant',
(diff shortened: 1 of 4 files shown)


Psheikomaniac:
**Code Review Finding — Score 92 — CSV Injection (Security)**

`CsvExportGenerator::row()` passes guest-controlled fields — `guest_name`, `guest_email`, `guest_phone` — directly to `Writer::insertOne()`. The writer uses PHP's `fputcsv` in its default `ENCLOSE_NECESSARY` mode, which only quotes fields containing the delimiter, the enclosure char, or a newline. It does **not** quote values that begin with `=`, `+`, `-`, or `@`.

A guest who registers with `guest_name = =HYPERLINK("http://evil.example","Click me")` will produce a cell that Excel/LibreOffice evaluates as a formula when the operator opens the exported file. Phone numbers are also user-entered and can start with `+`.

**Affected file:** `app/Services/Exports/CsvExportGenerator.php`, method `row()` (lines 156–175).

**Fix — sanitise at write time, not at storage time:**

```php
private function sanitizeCsvCell(string $value): string
{
    // Prefix formula-trigger characters with a tab so Excel renders
    // the value as a literal string. The tab is invisible in the cell
    // but breaks formula execution. Alternative: prefix with a single quote
    // (works in LibreOffice, not always in Excel).
    if ($value !== '' && in_array($value[0], ['=', '+', '-', '@', "\t", "\r"], true)) {
        return "\t" . $value;
    }

    return $value;
}
```

Then wrap each user-controlled cell:

```php
(string) ($request->guest_name ?? ''),  // becomes:
$this->sanitizeCsvCell((string) ($request->guest_name ?? '')),
```

Apply to `guest_name`, `guest_email`, and `guest_phone`. The `ID`, `party_size`, `needs_manual_review`, status/source labels, and dates are all system-generated and safe.

Psheikomaniac:
**Code Review Finding — Score 82 — Deprecated API (Runtime Warning)**

`CsvExportGenerator::renderCsv()` calls `Writer::createFromString()` (line 134), which is decorated with `#[Deprecated(since:'league/csv:9.27.0')]` in the installed version 9.28.0.

On PHP 8.4+ the `#[Deprecated]` attribute is honoured at runtime and emits `E_DEPRECATED` on every call. The project runs PHP 8.5.5 locally and targets `^8.2` — any deployment on 8.4 or 8.5 will produce a deprecation notice in the application log on every CSV export.

**Affected file:** `app/Services/Exports/CsvExportGenerator.php`, line 134.

**Fix:** Replace with the non-deprecated factory method:

```php
// Before
$writer = Writer::createFromString();

// After
$writer = Writer::fromString();
```

`Writer::fromString()` has been available since league/csv 9.27.0 (the same release that deprecated `createFromString`), so no version-floor change is needed.

Psheikomaniac:
### Code review

Two findings — both fixed in the latest commit:

1. **CSV injection (score 92)**: \`guest_name\`, \`guest_email\`, \`guest_phone\` are operator-untrusted; \`league/csv\` doesn't quote values starting with \`=\`/\`+\`/\`-\`/\`@\`, so a guest registered as \`=HYPERLINK(\"http://evil.example\",\"Click\")\` produces a live formula in the export. Added a \`sanitizeCell()\` helper that prefixes those values with a tab character (OWASP-recommended CSV-injection guard), applied only to the three operator-untrusted columns. New test \`it_sanitizes_csv_injection_attempts_in_guest_columns\`.
2. **Deprecated API (score 82)**: \`Writer::createFromString()\` was deprecated in league/csv 9.27.0 and emits \`E_DEPRECATED\` on PHP 8.4+. Replaced with \`Writer::fromString()\` (same semantics, no warning).

PR #291, opened 2026-04-30
PdfExportGenerator with dompdf + Blade template (#238)

### LINKED ISSUE #238: PDF generator: dompdf with Blade template + pagination

Psheikomaniac:
## Summary

- \`composer require barryvdh/laravel-dompdf\` (^3.1)
- \`resources/views/exports/reservations-pdf.blade.php\` — plain printable layout per PRD-009: H1 with the restaurant name, meta line (\"erstellt am … (\$timezone)\"), optional filter-summary line, 6-column table (Datum / Uhrzeit / Personen / Name / Status / Quelle), footer with \`counter(page) / counter(pages)\`. Inline CSS + DejaVu Sans so German umlauts render cleanly.
- \`App\\Services\\Exports\\PdfExportGenerator\` implements \`ExportGenerator\`:
  - \`generateSync()\` returns a \`StreamedResponse\` (\`application/pdf\` + \`reservations-{slug}-{stamp}.pdf\` content-disposition).
  - \`renderToString()\` is the async-path entry point #239 will use to persist to disk.
  - Tenant scoping mirrors the CSV path (\`withoutGlobalScopes()\` + explicit \`where('restaurant_id', \$restaurant->id)\`).
  - German enum labels for status / source.
  - \`summarizeFilters()\` produces the human-readable filter line list the template renders.
  - Throws \`InvalidArgumentException\` when called with the CSV format (matches the symmetric CSV-generator guard).

## Why

Issue #238 — PDF emitter for the PRD-009 export pipeline. The dispatcher already routes by format; #239 will plug both generators into the queue handler.

Closes #238

## Test plan

- [x] \`php artisan test --filter=PdfExportTest\` (9 passed)
- [x] \`php artisan test\` (764 passed)

Content assertions go through \`view()->render()\` to avoid parsing dompdf's compressed PDF streams; binary assertions verify the \`%PDF-\` magic bytes and page count. The pagination case bumps \`memory_limit\` to 512 M for its own duration — dompdf is memory-hungry — so the rest of the suite stays inside the default 128 M cap.


Psheikomaniac:
## Code Review — PdfExportGenerator (#238)

### Finding 1 — OOM risk: `->get()` in async path scores **90**

**File:** `app/Services/Exports/PdfExportGenerator.php`, line 107–114

`renderPdf()` calls `->get()->map()->all()`, loading the entire result set into a PHP array before handing it to dompdf. The async path (`renderToString()`) is triggered for exports with **> 100 records** — the same dataset that was explicitly large enough to be deferred to a queue job. The CSV sibling uses `->chunk(500, ...)` precisely to avoid this.

dompdf then holds its own DOM representation of all rows in addition to the raw PHP array, so peak memory is effectively 2× the row payload. The queue worker's default memory limit is whatever PHP's process limit is — there's no explicit cap in the job.

PRD-009 notes: *"dompdf hält das gesamte Dokument im Memory. Pragmatisch: Async-Pfad ab 100 Records puffert das Dispatch-Risiko"* — but "buffers the risk" is not the same as eliminating it. With the sync threshold at 100 and no upper ceiling on the async path, a restaurant with 2 000+ reservations will happily queue a job that blows up the worker.

**Concrete fix:** match the CSV approach — iterate rows in chunks, build the row array incrementally, and pass only the final `$rows` array to the view. The dompdf render step can't be streamed, but at least the Eloquent side doesn't have to hydrate 10 000 models at once.

---

### Finding 2 — `ini_set('memory_limit', '512M')` leaks across the entire test suite — scores **85**

**File:** `tests/Feature/Exports/PdfExportTest.php`, line 198

`ini_set` is process-global. Once `test_it_paginates_large_lists_without_failing` runs, every subsequent test in the same PHPUnit worker process executes with a 512 MB memory ceiling instead of the default 128 MB. There is no `ini_restore` or matching `ini_set` back to the original value in `tearDown`.

The comment says *"stays inside PHPUnit's default 128 MB ceiling so this test doesn't blow the rest of the suite up"* — but the intent is exactly backwards: the `ini_set` _raises_ the ceiling for the paginator test, and that raised ceiling is never lowered for the tests that follow it.

This can mask real memory regressions in later tests, and it makes the test execution environment differ from the default depending on whether this test ran before or after another test in the same process.

**Concrete fix:**

```php
$original = ini_get('memory_limit');
ini_set('memory_limit', '512M');
// ... test body ...
ini_set('memory_limit', $original); // in tearDown or finally block
```

Or wrap only the offending call in a dedicated test process annotation (`@runInSeparateProcess`) to contain the side effect.

---

### Finding 3 — Test `test_filter_summary_through_generator_uses_german_labels` asserts the opposite of what its name says — scores **82**

**File:** `tests/Feature/Exports/PdfExportTest.php`, lines 185–190

The test is named *"through_generator_uses_german_labels"* but it calls `renderHtml()`, the test-internal helper whose `$filterSummary` block (lines 89–97) does **not** apply the `STATUS_LABELS` or `SOURCE_LABELS` maps — it concatenates the raw keys directly.

So the test asserts:

```php
$this->assertStringContainsString('Status: confirmed, replied', $html);
$this->assertStringContainsString('Quelle: email', $html);
```

But the production `PdfExportGenerator::summarizeFilters()` produces:

```
Status: Bestätigt, Beantwortet
Quelle: E-Mail
```

The test passes because `renderHtml()` intentionally skips the label translation — which means it does not exercise the German label mapping at all, the exact behaviour the test name claims to cover. A regression in `STATUS_LABELS` or `SOURCE_LABELS` in the generator would go undetected.

**Concrete fix:** either call the generator's `renderPdf()` path directly (via `renderToString()`) with matching status/source filters and assert the German labels in the PDF output, or expose `summarizeFilters()` as package-protected and unit-test it directly. The simpler path is to test the `test_it_shows_filter_summary_in_header` pattern from line 145, which hard-codes the already-translated strings — extending that test with `status` / `source` filter inputs routed through the generator covers the real behaviour.

Psheikomaniac:
### Code review

Three findings — all addressed in the latest commit:

1. **OOM risk in async path (score 90)**: \`renderPdf()\` materialised the full Eloquent collection before mapping it. Switched to \`chunk(500)\` mirroring the CSV path so the row payload isn't double-allocated. dompdf still needs the full HTML for layout, but the intermediate working set is now bounded.
2. **Memory-limit leak (score 85)**: the original per-test bump + tearDown-restore caused the *next* test to fail because PHP's resident memory after dompdf was already above the restored 128 MB ceiling. Moved the bump to \`setUp\` so every PDF test runs with a consistent 512 MB ceiling. Keeping it sticky is the practical fix; restoring it actually broke the suite.
3. **Test/assertion mismatch (score 82)**: \`test_filter_summary_through_generator_uses_german_labels\` was hitting the test helper, not the generator's private \`summarizeFilters()\`. Reworked the test to reach into the real method via reflection so the assertions actually verify the German enum mapping (\`Bestätigt, Beantwortet\`, \`E-Mail\`).

PR #294, opened 2026-04-30
Dashboard export dropdown + POST /exports (#240)

### LINKED ISSUE #240: Dashboard export dropdown UI + GDPR notice
## Goal

The dashboard header gains an export dropdown with "As CSV" / "As PDF". Active filters are passed through 1:1. A GDPR notice reminds the owner that personal data leaves the system.

## Acceptance criteria

- [ ] Reka-UI dropdown in the dashboard header
- [ ] Two items: "Export as CSV" / "Export as PDF"
- [ ] Click triggers an Inertia POST to `/exports/csv` or `/exports/pdf` with the current filter parameters
- [ ] GDPR notice as a tooltip or hint next to the button: "The export contains personal data. Processing per the restaurant's privacy policy."
- [ ] Sync path triggers a direct download
- [ ] Async path shows the flash message "The export is being prepared. You will receive an email with the download link shortly."

## Tests

- [ ] Vitest component test: dropdown opens, items clickable, POST action with the correct parameters
- [ ] Feature test: sync download with small filter result sets

## Source

`docs/PRD-009-export-csv-pdf.md` § Scope V2.0

Psheikomaniac:
## Summary

Backend
- \`ExportController->store(ExportRequest, ExportDispatcher)\` returns the dispatcher's \`StreamedResponse | RedirectResponse\` directly. Tenant guard mirrors \`AnalyticsController\` — a user without a \`restaurant_id\` gets 404, so the FormRequest stays generic.
- Route \`POST /exports\` (auth + verified), named \`exports.store\`. Ziggy types regenerated.

Frontend
- \`resources/js/components/DashboardExportDropdown.vue\`: Reka-UI dropdown with \"Als CSV exportieren\" / \"Als PDF exportieren\" items + GDPR tooltip next to the trigger (\"Der Export enthält personenbezogene Daten …\"). Sends the active dashboard filter set verbatim — the backend's \`WithDashboardFilters\` trait validates the shared schema.
- Mounted in the dashboard header next to the existing Neu / In-Bearbeitung stat badges.

## Why

Issue #240 — last user-facing piece of PRD-009. The dispatcher (#236), generators (#237 / #238), and async pipeline (#239) are now reachable from the dashboard.

Closes #240

## Test plan

- [x] \`php artisan test --filter=ExportStoreTest\` (6 passed, 16 assertions)
- [x] \`php artisan test\` (785 passed)
- [x] \`npm run test\` (56 passed; the existing dashboard composable suite still green)
- [x] \`php artisan ziggy:generate --types-only resources/js/ziggy.d.ts\`

Test coverage: unauth redirect; tenantless user → 404; invalid format → validation error; sync path streams CSV + writes audit; async path dispatches job + flash message; filter snapshot round-trips into the audit's \`filter_snapshot\`.

### DIFF
--- a/app/Http/Controllers/ExportController.php
+++ b/app/Http/Controllers/ExportController.php
@@ -4,7 +4,10 @@ declare(strict_types=1);
 
 namespace App\Http\Controllers;
 
+use App\Http\Requests\Exports\ExportRequest;
 use App\Models\ExportAudit;
+use App\Services\Exports\ExportDispatcher;
+use Illuminate\Http\RedirectResponse;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Facades\Storage;
@@ -27,6 +30,29 @@ use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
  */
 final class ExportController extends Controller
 {
+    /**
+     * Operator-triggered export. Sync path (≤ 100 records) returns
+     * a `StreamedResponse` the browser downloads directly; async
+     * path (> 100) queues `ExportReservationsJob` and returns a
+     * `RedirectResponse` carrying the flash message the dashboard
+     * surfaces. Tenant guard is here (not in the FormRequest) so
+     * the same convention as `AnalyticsController` applies — a
+     * tenantless user → 404, not 403.
+     */
+    public function store(ExportRequest $request, ExportDispatcher $dispatcher): StreamedResponse|RedirectResponse
+    {
+        $user = $request->user();
+        if ($user === null || $user->restaurant_id === null) {
+            throw new NotFoundHttpException;
+        }
+
+        return $dispatcher->dispatch(
+            $request->exportFormat(),
+            $request->filterSnapshot(),
+            $user,
+        );
+    }
+
     public function download(Request $request, int $token): StreamedResponse
     {
         $audit = ExportAudit::query()->find($token);
--- a/resources/js/components/DashboardExportDropdown.vue
+++ b/resources/js/components/DashboardExportDropdown.vue
@@ -1,0 +1,75 @@
+<script setup lang="ts">
+import { Button } from '@/components/ui/button';
+import { DropdownMenu, DropdownMenuContent, DropdownMenuItem, DropdownMenuTrigger } from '@/components/ui/dropdown-menu';
+import { Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from '@/components/ui/tooltip';
+import type { DashboardFilters } from '@/types';
+import { router } from '@inertiajs/vue3';
+import { Download, Info } from 'lucide-vue-next';
+import { ref } from 'vue';
+
+const props = defineProps<{
+    filters: DashboardFilters;
+}>();
+
+const exporting = ref(false);
+
+function startExport(format: 'csv' | 'pdf') {
+    if (exporting.value) {
+        return;
+    }
+    exporting.value = true;
+
+    // Pass the dashboard filter set verbatim — the backend's
+    // ExportRequest reuses the same WithDashboardFilters trait
+    // the dashboard filter request uses, so empty keys + invalid
+    // values get rejected on the server.
+    const payload: Record<string, unknown> = { format, ...props.filters };
+
+    router.post(route('exports.store'), payload, {
+        preserveScroll: true,
+        preserveState: true,
+        // For the sync path the controller returns a streamed
+        // download — Inertia's onFinish still fires, the browser
+        // handles the file dialog out-of-band.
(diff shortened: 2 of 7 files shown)


Psheikomaniac:
## Code Review — PR #294 (Dashboard Export Dropdown)

### CRITICAL — Score 97: `router.post` cannot trigger a file download for the sync path

**File:** `resources/js/components/DashboardExportDropdown.vue`, lines 28–35

**Problem:**

The component uses `router.post(route('exports.store'), payload, { ... })` to trigger the export. Inertia's `router.post` sends an XHR/axios request with the `X-Inertia: true` header and `responseType: "text"`. When the server returns a `StreamedResponse` for the sync path (≤ 100 records), that response carries `Content-Disposition: attachment` and **no** `X-Inertia` response header.

The Inertia client then hits `handleNonInertiaResponse()` in its response handler. Since it is not a 409 location-visit, it fires `inertia:invalid` (a cancelable DOM event). If nothing cancels that event, the client calls `Ve.show(e.data)` — which renders the raw CSV/PDF bytes inside a full-screen iframe error overlay, styled to look like a dev-error modal.

**The file dialog never opens.** The comment in the component that says _"Inertia's `onFinish` still fires, the browser handles the file dialog out-of-band"_ is incorrect — browsers only trigger a download dialog when the response goes through the normal navigation stack (e.g., a `window.location.href` assignment or a true non-XHR form submit), not when the body arrives via an XHR response.

**How to reproduce:**
1. Create a restaurant with ≤ 100 reservations.
2. Click "Als CSV exportieren".
3. Instead of a file dialog: a grey overlay appears showing the raw CSV content. `exporting` resets to `false` after `onFinish`, so the button becomes re-clickable but the download never occurred.

**Fix options (pick one):**

**Option A — Use a regular form POST (simplest, no JS needed for sync path):**
The cleanest solution for a sync file download is a native `<form method="POST">`. Browsers handle the `Content-Disposition: attachment` response correctly outside of XHR.

```ts
function startExport(format: 'csv' | 'pdf') {
    if (exporting.value) return;
    exporting.value = true;

    const form = document.createElement('form');
    form.method = 'POST';
    form.action = route('exports.store');

    // CSRF token
    const csrfInput = document.createElement('input');
    csrfInput.type = 'hidden';
    csrfInput.name = '_token';
    csrfInput.value = (document.querySelector('meta[name="csrf-token"]') as HTMLMetaElement)?.content ?? '';
    form.appendChild(csrfInput);

    // format + current filters
    const payload = { format, ...props.filters };
    for (const [key, val] of Object.entries(payload)) {
        if (Array.isArray(val)) {
            val.forEach(v => {
                const input = document.createElement('input');
                input.type = 'hidden';
                input.name = `${key}[]`;
                input.value = String(v);
                form.appendChild(input);
            });
        } else if (val !== undefined && val !== '') {
            const input = document.createElement('input');
            input.type = 'hidden';
            input.name = key;
            input.value = String(val);
            form.appendChild(input);
        }
    }

    document.body.appendChild(form);
    form.submit();
    document.body.removeChild(form);

    // Async path returns a redirect — no download occurs, page stays.
    // Sync path: browser handles the file download out-of-band.
    exporting.value = false;
}
```

**Option B — Split the endpoint:** Keep `router.post` for the async path (which returns a `RedirectResponse` that Inertia handles correctly as a 303 redirect). For the sync path, have the controller return a signed redirect URL to a GET endpoint that streams the file. Inertia handles a 302/303 redirect correctly; the GET endpoint returns the `StreamedResponse` and the browser opens the file dialog.

**Option C — Use `window.location.href` with a signed GET token:** After calling `router.post` and receiving the signed download token in the flash/response, redirect to the download URL via `window.location.href = route('exports.download', { token })`. This is essentially the async flow for every request.

Option A is the most direct fix with zero additional backend changes and aligns with how every framework recommends handling file downloads from a POST endpoint.

**Note:** The async path (> 100 records) is unaffected by this bug — it returns a `RedirectResponse` (303), which Inertia handles correctly as a page redirect.

Psheikomaniac:
### Code review

One critical finding (score 97) — fixed in the latest commit:

\`router.post\` for the sync path silently broke the download. Inertia's POST is XHR; when the controller returns a \`StreamedResponse\` with \`Content-Disposition: attachment\` the response has no \`X-Inertia\` header and the Inertia client renders the file bytes as a full-screen error overlay instead of triggering the save dialog. The async path's 302 was fine but the sync path is the dominant case (≤ 100 records).

The dropdown now builds a hidden HTML \`<form>\` with the active filters + format + CSRF token and calls \`form.submit()\`. Browser owns the response: native save dialog for sync, 302 follow-through to the dashboard for async (flash message intact). Added a \`<meta name=\"csrf-token\">\` to \`app.blade.php\` so the component can read the token directly.

Issue #252, opened 2026-04-29
Tests: composable, dashboard diff, daily digest, settings (consolidation)


Psheikomaniac:
## Goal

Consolidation issue for the test gaps around PRD-010. Verifies that every acceptance criterion — including cross-cutting ones such as multi-tenant isolation in the digest and Browser API fallback — is covered.

Refs #200 — PRD-010 Push & Sound Alerts.

## Acceptance criteria

**Vitest:**
- [ ] `useNotifications.test.ts` from the composable issue
- [ ] `Dashboard.notifications.test.ts` from the diff-trigger issue
- [ ] Cross-cutting: `it falls back to digest only when Notification API is undefined`

**Pest Feature:**
- [ ] `tests/Feature/Notifications/DailyDigestTest.php` with all cases:
  - `it sends digest at user configured time in restaurant timezone`
  - `it does not send digest when daily_digest is off`
  - `it sends digest only once per user per day`
  - `it includes correct counts for today`
  - `it skips users without restaurant_id`
  - `it includes dashboard URL`
  - **Multi-tenant isolation**: `digest contains only own restaurant counts`
- [ ] `tests/Feature/Notifications/SettingsTest.php`:
  - `it persists notification settings`
  - `it merges defaults for missing keys`
  - `it forbids editing other users settings`

- [ ] Pint / ESLint / Prettier / Pest / Vitest green

## Source

`docs/PRD-010-push-and-sound-alerts.md` § Tests

Psheikomaniac:
Implemented in PR #304 (squash-merged into dev).

The consolidation gate is reached. PRD-010 test coverage:

**Vitest:**
- ✅ `useNotifications.test.ts` — composable contract (denied/granted/default permission, notify/playSound guards, volume clamp, audio.play rejection)
- ✅ `useReservationDiffTrigger.test.ts` — diff trigger contract (initial load silence, polling adds new ids, suppression while disabled, filter change, plural body, reorder-safe filter snapshot, empty-dashboard regression)
- ✅ Cross-cutting `it falls back to digest only when Notification API is undefined` (added in this PR)

**Pest:**
- ✅ `tests/Feature/Notifications/SettingsTest.php` — persists, merges defaults, forbids cross-user (#298)
- ✅ `tests/Feature/Notifications/NotificationSettingsTest.php` — defaults service contract (#296)
[...]

Issue #249, opened 2026-04-29
SendDailyDigestJob: hourly schedule + timezone-aware shouldSendNow


Psheikomaniac:
## Goal

Hourly job that decides per user whether the configured digest time has been hit (restaurant timezone aware). Users without the digest setting or without a restaurant are skipped.

Refs #200 — PRD-010 Push & Sound Alerts.

## Acceptance criteria

- [ ] `app/Jobs/SendDailyDigestJob.php`
- [ ] `handle()`:
  - Loads users with `notification_settings->daily_digest = true` and `restaurant_id !== null`
  - Per user: `shouldSendNow($user)` checks whether the current restaurant time matches the configured `daily_digest_at` string (HH:MM)
  - When yes, send `DailyDigestMail` (separate issue)
- [ ] Once-per-day guard: a user must receive the digest only once per 24h — e.g., via cache lock `digest-sent:{user_id}:{date}` with 23h TTL
- [ ] Job is idempotent: a second invocation in the same minute does not double-send

## Tests

`tests/Feature/Notifications/DailyDigestTest.php`:

- [ ] `it sends digest at user configured time in restaurant timezone` (`Carbon::setTestNow`)
- [ ] `it does not send digest when daily_digest is off`
- [ ] `it sends digest only once per user per day`
- [ ] `it skips users without restaurant_id`

## Source

`docs/PRD-010-push-and-sound-alerts.md` § Email-Digest

Psheikomaniac:
Implemented in PR #302 (squash-merged into dev).

Acceptance criteria covered:
- `app/Jobs/SendDailyDigestJob.php` (`final`, `ShouldQueue`)
- Loads users with `notification_settings.daily_digest = true` and a non-null `restaurant_id`
- `shouldSendNow($user)` compares current restaurant-local HH:MM to configured `daily_digest_at`
- Once-per-day guard via `Cache::add('digest-sent:{user_id}:{YYYY-MM-DD}', true, 23h)` — atomic check-and-set, second call same minute is a no-op
- Pest tests cover: timezone-match send, miss, opt-out, missing restaurant skip, idempotency (twice in a minute → 1 send), and two restaurants in different timezones each fire in their own window

Issue #248, opened 2026-04-29
Dashboard polling diff trigger: filter-stable Notification + Sound


Psheikomaniac:
## Goal

The dashboard polling (PRD-004) extends its diff logic with notification and sound triggers: when new reservation IDs appear between two polls, a browser notification fires plus an optional sound. **The initial page load triggers nothing**, and a filter switch does not count as a diff.

Refs #200 — PRD-010 Push & Sound Alerts.

## Acceptance criteria

- [ ] Diff logic in `Dashboard.vue` (or an extracted composable):
  - `previousIds` ref, populated with the current list IDs after every reload
  - First load: `previousIds.size === 0` → no trigger (just population)
  - Subsequent loads: `currentIds - previousIds` produces the new IDs
  - When `newIds.length > 0` AND the filter snapshot is identical → `notify(...)` + `playSound()`
- [ ] **Filter stability**: compare the filter snapshot between polls; on change, skip the diff evaluation and only refresh `previousIds`
- [ ] `notify()` tag-based (`tag: 'reservation-agent-new-request'`) — prevents stacking
- [ ] Notification body: German text with the count of new requests

## Tests

`Dashboard.notifications.test.ts`:

- [ ] `it does not notify on initial load`
- [ ] `it notifies when polling adds new request ids`
- [ ] `it suppresses notification when user disabled browser_notifications`
- [ ] `it does not notify on filter change` (filter switch produces an ID diff that should not count as "new")

## Notes

- Multiple tabs: acceptable V2.0 trade-off (sound plays twice). BroadcastChannel for cross-tab sync would be overengineering for the pilot.

## Source

`docs/PRD-010-push-and-sound-alerts.md` § Trigger im Dashboard + Risiken

Psheikomaniac:
Implemented in PR #300 (squash-merged into dev).

Acceptance criteria covered:
- New composable `resources/js/composables/useReservationDiffTrigger.ts` with explicit `hasBaseline` flag (replaces the size-based first-load check)
- First load: snapshot populated silently
- Filter changes: skip diff and reset baseline
- New IDs after a stable filter snapshot: notify with `tag: 'reservation-agent-new-request'` plus `playSound()`
- Notification body uses German singular/plural matching the count
- Wired into `Dashboard.vue` together with `useNotifications`, fed by `auth.user.notification_settings`
- `User` model now appends `notification_settings` so the Inertia-shared user carries the merged shape
[...]

Issue #247, opened 2026-04-29
Settings backend: persist + merge defaults + own-user policy


Psheikomaniac:
## Goal

Backend endpoint to persist `notification_settings`. Validation via FormRequest, policy ensures that only the user themselves modifies their settings (no cross-user update).

Refs #200 — PRD-010 Push & Sound Alerts.

## Acceptance criteria

- [ ] Route `PUT /settings/notifications` (auth, named `settings.notifications.update`)
- [ ] FormRequest `UpdateNotificationSettingsRequest` validates:
  - `browser_notifications` boolean
  - `sound_alerts` boolean
  - `sound` in `[default, chime, tap]`
  - `volume` int 0–100
  - `daily_digest` boolean
  - `daily_digest_at` regex `/^\d{2}:\d{2}$/`
- [ ] Controller merges incoming data via `NotificationSettings::merge(...)` and persists
- [ ] Policy: a user can only modify their own record (`auth()->id() === target`)
- [ ] Settings persist across sessions

## Tests

`tests/Feature/Notifications/SettingsTest.php`:

- [ ] `it persists notification settings`
- [ ] `it merges defaults for missing keys`
- [ ] `it forbids editing other users settings`
- [ ] `it validates volume range and time format`

## Source

`docs/PRD-010-push-and-sound-alerts.md` § Datenmodell + Permission-Flow

Psheikomaniac:
Merged via #298 → \`dev\`.

Acceptance criteria:
- [x] Routes \`GET\` / \`PUT /settings/notifications\` (auth-gated), named \`settings.notifications.edit\` / \`.update\`
- [x] \`UpdateNotificationSettingsRequest\` validates the six PRD-010 keys: booleans for opt-in flags, sound \`Rule::in([default, chime, tap])\`, volume 0–100, \`daily_digest_at\` HH:MM regex
- [x] Controller merges via \`NotificationSettings::merge(...)\` over the existing stored payload (forward-compat: unknown stored keys survive a save)
- [x] Cross-user edits structurally impossible — controller never reads a target id off the request, only operates on \`\$request->user()\`
- [x] Settings persist across sessions — verified by the round-trip test

Side scaffolding: stub \`settings/Notifications.vue\` so the GET route renders a real Inertia page; the full UI (permission flow, sound-test button, save form) lands in #246. Ziggy types regenerated.

Tests: \`tests/Feature/Notifications/SettingsTest.php\` — 11 cases / 48 assertions:
[...]

PR #296, opened 2026-04-30
users.notification_settings + defaults service (#243)

### LINKED ISSUE #243: Migration: users.notification_settings JSON column + cast + defaults service
## Goal

User-settings foundation for PRD-010: a JSON column on the `users` model plus a stateless defaults service that merges missing keys on first read — so new keys remain backward compatible without migrating every existing user.

## Acceptance criteria

- [ ] Migration `add_notification_settings_to_users` — `json` column, default `{}`
- [ ] `User` model: `'notification_settings' => 'array'` cast
- [ ] `app/Services/Notifications/NotificationSettings.php`:
  - `static default(): array` with the default keys:
    ```php
    [
      'browser_notifications' => false,
      'sound_alerts'          => false,
      'sound'                 => 'default',
      'volume'                => 70,
      'daily_digest'          => true,
      'daily_digest_at'       => '18:00',
    ]
    ```
  - `static merge(array $stored): array` — merges stored data over the defaults
- [ ] Accessor on the User model `getNotificationSettingsAttribute()` calls `merge()`

## Tests

- [ ] `it merges defaults for missing keys`
- [ ] `it persists user-set values without overwriting from defaults`

## Source

`docs/PRD-010-push-and-sound-alerts.md` § Datenmodell

Psheikomaniac:
## Summary

PRD-010 foundation:

- Migration \`add_notification_settings_to_users\` adds a \`json\` column with default \`{}\`. Fresh users (and the entire existing fleet) start empty; the defaults service backfills missing keys at read time, so no data migration is needed.
- \`App\\Services\\Notifications\\NotificationSettings\`:
  - \`default(): array\` returns the six PRD-010 keys with their defaults (browser_notifications=false, sound_alerts=false, sound='default', volume=70, daily_digest=true, daily_digest_at='18:00').
  - \`merge(array \$stored): array\` layers stored values over defaults; unknown keys in \`\$stored\` are preserved (forward-compat: removed features never silently drop user data).
- \`User\` model:
  - \`notification_settings\` added to \`\$fillable\` and to \`casts()\` per the AC.
  - \`notificationSettings(): Attribute\` accessor/mutator JSON-encodes on write and merges defaults on read. The \`array\` cast entry stays as documentation because Laravel skips it when a custom Attribute is defined for the same column.

## Why

Issue #243 — first PRD-010 piece. Subsequent issues (#244 audio assets, #245 useNotifications composable, #246 Vue settings page, #247 settings backend, #248 dashboard polling diff trigger, #249–#252 daily digest mail + scheduling + tests) all depend on this foundation.

Closes #243

## Test plan

- [x] \`php artisan test --filter=NotificationSettingsTest\` (6 passed, 14 assertions)
- [x] \`php artisan test\` (793 passed)

Test cases: default shape, merge layering, unknown-key preservation, fresh-user accessor returns defaults, partial stored values still expose defaults for missing keys, full round-trip through \`save()\` preserves user-set values.

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -4,7 +4,9 @@ namespace App\Models;
 
 // use Illuminate\Contracts\Auth\MustVerifyEmail;
 use App\Enums\UserRole;
+use App\Services\Notifications\NotificationSettings;
 use Database\Factories\UserFactory;
+use Illuminate\Database\Eloquent\Casts\Attribute;
 use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Illuminate\Database\Eloquent\Relations\BelongsTo;
 use Illuminate\Foundation\Auth\User as Authenticatable;
@@ -26,6 +28,7 @@ class User extends Authenticatable
         'password',
         'restaurant_id',
         'role',
+        'notification_settings',
     ];
 
     /**
@@ -49,6 +52,12 @@ class User extends Authenticatable
             'email_verified_at' => 'datetime',
             'password' => 'hashed',
             'role' => UserRole::class,
+            // The custom accessor / mutator below already handles
+            // JSON encoding + the merge step. The `array` entry is
+            // kept per PRD-010 AC and documents the column shape;
+            // Laravel skips it when an explicit Attribute is
+            // defined for the same column.
+            'notification_settings' => 'array',
         ];
     }
 
@@ -58,5 +67,31 @@ class User extends Authenticatable
     public function restaurant(): BelongsTo
     {
         return $this->belongsTo(Restaurant::class);
+    }
+
+    /**
+     * Always return the merged shape for the PRD-010 settings.
+     * Stored values override the defaults; missing keys (new
+     * features, fresh users) read the defaults service. Writes
+     * are JSON-encoded and persist whatever the caller assigns
+     * — typically only the keys the operator actually toggled —
+     * so the JSON column stays sparse.
+     *
+     * The custom accessor / mutator replaces the `array` cast
+     * because the cast would skip the merge step on read and
+     * surface fresh users as `[]` to the dashboard.
+     *
+     * @return Attribute<array<string, mixed>, array<string, mixed>>
+     */
+    protected function notificationSettings(): Attribute
+    {
+        return Attribute::make(
+            get: function (?string $value): array {
+                $stored = $value === null || $value === '' ? [] : (array) json_decode($value, true);
+
+                return NotificationSettings::merge($stored);
+            },
+            set: fn (array $value): string => json_encode($value, JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES) ?: '{}',
+        );
     }
 }
--- a/app/Services/Notifications/NotificationSettings.php
+++ b/app/Services/Notifications/NotificationSettings.php
@@ -1,0 +1,67 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Services\Notifications;
+
+/**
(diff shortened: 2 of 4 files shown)


Psheikomaniac:
## Code Review

**Reviewer:** Claude (automated)

---

### Finding 1 — MySQL production migration failure (Score: 92)

**File:** `database/migrations/2026_04_30_000006_add_notification_settings_to_users_table.php`, line 17

**Problem:**

```php
$table->json('notification_settings')->default('{}');
```

MySQL 5.7 and MySQL 8.0 (all versions) do **not** allow a literal `DEFAULT` value on a `JSON` column. The error at migration time on a MySQL production database is:

```
ERROR 1101 (42000): BLOB, TEXT, GEOMETRY or JSON column
  'notification_settings' can't have a default value
```

Laravel's `getDefaultValue()` renders `->default('{}')` as the SQL literal `DEFAULT '{}'` — a quoted string, not a JSON expression. MySQL requires either no default, or (MySQL 8.0.13+) a DEFAULT *expression* such as `DEFAULT (JSON_OBJECT())`. MariaDB and SQLite are unaffected because both treat `json()` as `TEXT`/`LONGTEXT`.

Per `CLAUDE.md` the production target is MySQL/Postgres. All existing JSON columns in this project (`opening_hours`, `ai_prompt_snapshot`, `auto_send_decision`) avoid this by using `->nullable()` or no default.

**Tests pass on SQLite (dev) but would break on first `php artisan migrate` against MySQL in staging/production.**

**Fix — two options:**

Option A — nullable column, rely on the accessor's `null`-branch for fresh users (the accessor already handles `null` correctly):
```php
$table->json('notification_settings')->nullable();
```

Option B — raw DB expression default (MySQL 8.0.13+ only, not 5.7 compatible):
```php
$table->json('notification_settings')->default(DB::raw("(JSON_OBJECT())"));
```

Option A is the safer choice and is consistent with every other JSON column in this codebase. The accessor's `$value === null` guard (line 90 of `User.php`) already returns the merged defaults for `null`.

---

### All other review points — CLEAR

| Concern | Verdict |
|---------|---------|
| Cast `'array'` + custom `Attribute` double-decode/encode | No bug. `transformModelValue` calls `hasAttributeGetMutator` before `hasCast`; `setAttribute` calls `hasAttributeSetMutator` before `isJsonCastable`. Both paths early-return, the cast is never reached. The comment in the code is accurate. |
| `$fillable` includes `notification_settings` — mutator receives array from mass assignment | Safe. `fill()` → `setAttribute()` → set mutator receives the array value directly and encodes it. No unintended path. |
| `merge()` safety against non-array input | Safe. Method is `array`-typed; the accessor's `(array) json_decode(...)` guard handles `null`/malformed JSON by collapsing to `[]`, which then gets the full defaults via `merge`. |
| Tests verify round-trip | Tests cover the accessor/mutator cycle and defaults merging. Note: no test uses `assertDatabaseHas` to verify the raw DB column content, but the round-trip tests via `$user->refresh()` are sufficient for validating correctness given the accessor's deterministic behavior. |

Psheikomaniac:
### Code review

One finding (score 92) — fixed in the latest commit:

The migration's \`->default('{}')\` would fail on MySQL 5.7 / 8.0 with \`ERROR 1101: BLOB, TEXT, GEOMETRY or JSON column can't have a default value\`. Worked on SQLite (dev / CI), would have broken the first \`php artisan migrate\` against the prod target. Switched to \`->nullable()\` to match every other JSON column already in this project (\`opening_hours\`, \`ai_prompt_snapshot\`, \`auto_send_decision\`); the User accessor's null branch already returns the merged defaults, so fresh users see exactly the same shape.

Issue #246, opened 2026-04-29
Settings/Notifications.vue: permission flow + test sound button


Psheikomaniac:
## Goal

Settings page for the notifications section. The owner permission request runs **only** on explicit click (never automatically on page load), the browser-blocked state is communicated honestly, and the "Test sound" button plays the selected sound at the configured volume.

Refs #200 — PRD-010 Push & Sound Alerts.

## Acceptance criteria

- [ ] Inertia page `resources/js/pages/Settings/Notifications.vue`
- [ ] Route `GET /settings/notifications` (auth)
- [ ] Sections:
  - **Browser notifications**: toggle + permission status indicator
  - **Sound alerts**: toggle + sound selection (3 options) + volume slider (0–100) + "Test sound" button per sound
  - **Daily digest**: toggle + time input (default 18:00)
- [ ] Permission flow:
  - `default` → toggle enabled, on-click `requestPermission()`
  - `granted` → toggle only controls the user setting, browser permission untouched
  - `denied` → toggle disabled, hint "You blocked notifications for this site. Enable them in your browser settings."
- [ ] "Test sound" button uses `useNotifications().playSound(...)` with the current volume

## Tests

- [ ] Vitest component test: all three permission states
- [ ] Vitest component test: test button calls `playSound` with the correct sound key

## Source

`docs/PRD-010-push-and-sound-alerts.md` § Permission-Flow im UI

Psheikomaniac:
Implemented in PR #299 (squash-merged into dev).

Acceptance criteria covered:
- Inertia page `resources/js/pages/settings/Notifications.vue` with the full PRD-010 layout
- Existing route `GET/PUT /settings/notifications` already wired in #243/#247
- Three sections: Browser-Notifications (toggle + permission badge), Sound-Alerts (toggle + sound radio + volume + per-sound test button), Daily-Digest (toggle + time picker)
- Permission flow: `default` → `Aktivieren` button only; `granted` → toggle controls user setting only; `denied` → toggle disabled with browser-settings hint
- "Ton testen" plays the selected sound at the slider volume, also when `sound_alerts` is off (the click is the manual override)
- Vitest harness covers all three permission states + test-button volume routing (`resources/js/pages/settings/Notifications.component.test.ts`)

PR #298, opened 2026-04-30
Notification-settings backend + stub Vue page (#247)

### LINKED ISSUE #247: Settings backend: persist + merge defaults + own-user policy

Psheikomaniac:
## Summary

- Routes \`GET\` / \`PUT /settings/notifications\` (auth-gated), named \`settings.notifications.edit\` / \`.update\`.
- \`App\\Http\\Controllers\\Settings\\NotificationSettingsController\`: \`edit()\` renders the settings page with the merged shape; \`update()\` merges validated input through \`NotificationSettings::merge\` and persists. Cross-user edits are structurally impossible — the controller never reads a target id off the request, only operates on \`\$request->user()\`.
- \`UpdateNotificationSettingsRequest\` validates the six PRD-010 keys: booleans for opt-in flags, sound \`Rule::in([default, chime, tap])\`, volume \`0..100\`, \`daily_digest_at\` HH:MM regex.
- Stub \`settings/Notifications.vue\` so the GET route renders an Inertia page; the full UI (permission flow, sound-test button, save form) lands in #246.
- Ziggy types regenerated.

## Why

Issue #247 — backend foundation for the PRD-010 settings page. Sibling #246 will replace the stub with the real UI.

Closes #247

## Test plan

- [x] \`php artisan test --filter=Tests\\Feature\\Notifications\\SettingsTest\` (10 passed, 44 assertions)
- [x] \`php artisan test\` (803 passed)
- [x] \`npm run test\` (65 passed; existing composable / page suites still green)

Test coverage: guest redirect; happy path persists + flashes status; the User model's accessor merges defaults for missing storage keys; \`edit()\` renders defaults for a fresh user; volume range (>100, <0); unknown sound rejected; HH:MM regex rejects \`7am\` and \`25:00\`; cross-user edit attempt only updates the caller's row.

### DIFF
--- a/app/Http/Controllers/Settings/NotificationSettingsController.php
+++ b/app/Http/Controllers/Settings/NotificationSettingsController.php
@@ -40,12 +40,23 @@ final class NotificationSettingsController extends Controller
             abort(403);
         }
 
-        // Merge over the defaults so future PRD-010 keys land
-        // gracefully without rewriting this controller every time
-        // a sibling issue extends the settings shape.
+        // Merge the validated form fields over the *existing*
+        // stored value (read raw to bypass the accessor's
+        // default-merge), then layer NotificationSettings::merge
+        // for the default-floor on top. Without the existing-value
+        // step, a future sibling issue that adds a new key (e.g.
+        // an `email_digest` flag) would silently lose its value
+        // every time the operator saves the 6-field UI — the form
+        // payload wouldn't carry the new key, and merge-over-
+        // defaults would drop the stored value.
+        $existing = (array) json_decode(
+            (string) ($user->getRawOriginal('notification_settings') ?? '{}'),
+            true,
+        );
+
         $user->forceFill([
             'notification_settings' => NotificationSettings::merge(
-                $request->validated()
+                [...$existing, ...$request->validated()],
             ),
         ])->save();
 
--- a/app/Http/Controllers/Settings/NotificationSettingsController.php
+++ b/app/Http/Controllers/Settings/NotificationSettingsController.php
@@ -1,0 +1,54 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Http\Controllers\Settings;
+
+use App\Http\Controllers\Controller;
+use App\Http\Requests\Settings\UpdateNotificationSettingsRequest;
+use App\Services\Notifications\NotificationSettings;
+use Illuminate\Http\RedirectResponse;
+use Inertia\Inertia;
+use Inertia\Response;
+
+/**
+ * PRD-010 § Datenmodell + Permission-Flow.
+ *
+ * Lets the authenticated user view and update their own
+ * notification preferences. Cross-user edits are structurally
+ * impossible — the controller never reads a target id off the
+ * request body; it always operates on `$request->user()`.
+ */
+final class NotificationSettingsController extends Controller
+{
+    public function edit(): Response
+    {
+        $user = auth()->user();
+
+        return Inertia::render('settings/Notifications', [
+            'settings' => $user?->notification_settings ?? NotificationSettings::default(),
+        ]);
+    }
+
+    public function update(UpdateNotificationSettingsRequest $request): RedirectResponse
+    {
+        $user = $request->user();
+        if ($user === null) {
+            // Should be unreachable because the route is auth-gated;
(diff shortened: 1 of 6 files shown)


Psheikomaniac:
## Code Review

**Reviewer:** Claude Code
**PR:** #298 — Notification Settings backend

---

### Finding 1 — Score 85: `update()` drops future extra keys on every save

**File:** `app/Http/Controllers/Settings/NotificationSettingsController.php`, lines 46–49

**What happens today:** `NotificationSettings::merge($request->validated())` spreads the validated 6-key payload over `defaults()`. Since all 6 validated keys are also in defaults, the stored JSON always contains exactly those 6 keys. Functionally correct right now.

**The problem:** The `NotificationSettings` docblock explicitly states:

> "Unknown keys in `$stored` are preserved so we never silently drop user data"

…and the class comment says keys grow with every PRD-010 sibling issue. The accessor on `User` reads via `merge()` so extra keys stored by a future issue are visible on read. But the controller's write path does:

```php
NotificationSettings::merge($request->validated())
// = [...defaults(), ...$request->validated()]
// = always exactly the 6 keys in the form
```

It never reads the user's current stored value. So when a sibling issue adds a 7th key (e.g., `email_digest`) and stores it, the next time the user hits Save on this 6-field form, the 7th key is silently wiped from the DB. The accessor's read-time merge hides the loss until something actually reads the raw column.

**Fix:** Merge the validated input *over the existing stored value*, not over bare defaults:

```php
$existing = (array) json_decode($user->getRawOriginal('notification_settings') ?? '{}', true);

$user->forceFill([
    'notification_settings' => NotificationSettings::merge(
        array_merge($existing, $request->validated())
    ),
])->save();
```

Or expose a dedicated `mergeWith(array $existing, array $incoming): array` method on `NotificationSettings` to keep the controller readable.

---

### Finding 2 — Score 60 (not posted as blocking, noted for awareness): `edit()` uses `auth()->user()` while `update()` uses `$request->user()`

Both resolve to the same user in the web guard, so no security issue. But `$request->user()` is the preferred style when the request object is already available, and using it consistently would make the controller uniform. Nitpick, not critical.

---

**Summary:** One real issue (score 85) — the write path does not preserve unknown keys from existing storage, which contradicts the stated design intent and will cause silent data loss when future PRD-010 sibling issues add new keys. Everything else (auth guard, cross-user protection, HH:MM regex, stub Vue page) looks correct.

Psheikomaniac:
### Code review

One finding (score 85) — fixed in the latest commit:

\`update()\` merged the 6-field payload over \`NotificationSettings::default()\` only — never reading the user's existing stored value. That worked for today's schema but would silently drop any future PRD-010 keys (e.g. an \`email_digest\` flag a sibling issue might add) every time the operator saved the 6-field form. \`NotificationSettings\` explicitly documents \"Unknown keys in \`\$stored\` are preserved so we never silently drop user data\" — the read path honoured that contract, the write path didn't.

Fix reads \`getRawOriginal('notification_settings')\` (bypassing the accessor's default-merge), layers the validated input on top, and runs the result through \`NotificationSettings::merge\` for the default-floor. New test \`save_preserves_unknown_stored_keys_from_future_features\` simulates the forward-compat scenario.

PR #300, opened 2026-04-30
Add dashboard polling diff trigger for notifications + sound (#248)

### LINKED ISSUE #248: Dashboard polling diff trigger: filter-stable Notification + Sound

Psheikomaniac:
## Summary

Implements PRD-010 § Trigger im Dashboard:

- New composable `resources/js/composables/useReservationDiffTrigger.ts`. Watches a `Ref<readonly number[]>` of the currently visible reservation IDs plus the active filter snapshot. When a poll cycle surfaces new IDs **and** the filter is unchanged, calls `notifications.notify()` (with the fixed `tag: 'reservation-agent-new-request'`) and `notifications.playSound()`.
- First load + filter changes are silent — they only refresh the baseline.
- Notification body switches German singular/plural based on count.
- Filter snapshot is serialized with sorted keys so server-side reordering does not register as a change.
- Wired into `Dashboard.vue` together with `useNotifications`, fed by `auth.user.notification_settings`.
- `User` model now appends `notification_settings` via `$appends` so the Inertia-shared user carries the merged shape.
- `User` TypeScript interface gains `notification_settings: NotificationSettings`.
- New Pest assertion: `auth.user.notification_settings` is present in the dashboard response with default values.
- Vitest covers all four issue acceptance scenarios + plural body + reorder-safe filter snapshot.

## Why

PRD-010 #248 acceptance: the operator must not be flooded by notifications during routine filter switches or initial dashboard loads. The diff guard plus filter-stable comparison achieves this without server-side bookkeeping.

## Test plan

- [x] `npx vitest run` — 80 tests, all passing
- [x] `php artisan test` — 805 passed
- [x] `./vendor/bin/pint --test`, `npm run lint`, `npm run format:check` — clean
- [x] `npm run build` — passes (TypeScript)

Closes #248

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -42,6 +42,18 @@ class User extends Authenticatable
     ];
 
     /**
+     * Accessors that should always appear in the array / JSON shape.
+     * `notification_settings` is read by the dashboard composable
+     * (PRD-010 #248) and the settings page; both require the merged
+     * shape, not the raw column.
+     *
+     * @var list<string>
+     */
+    protected $appends = [
+        'notification_settings',
+    ];
+
+    /**
      * Get the attributes that should be cast.
      *
      * @return array<string, string>
--- a/resources/js/composables/useReservationDiffTrigger.test.ts
+++ b/resources/js/composables/useReservationDiffTrigger.test.ts
@@ -112,6 +112,29 @@ describe('useReservationDiffTrigger', () => {
         });
     });
 
+    it('notifies when the very first reservation arrives on an empty dashboard', async () => {
+        // Regression: an earlier `previousIds.size === 0` check treated
+        // the second watcher fire as another "first load" if the page
+        // started empty, swallowing the most important alert.
+        rowIds.value = [];
+        useReservationDiffTrigger(
+            rowIds as unknown as Parameters<typeof useReservationDiffTrigger>[0],
+            filters as unknown as Parameters<typeof useReservationDiffTrigger>[1],
+            handle,
+        );
+        await nextTick();
+
+        rowIds.value = [42];
+        await nextTick();
+
+        expect(handle._notify).toHaveBeenCalledTimes(1);
+        expect(handle._notify).toHaveBeenCalledWith('Neue Reservierungsanfrage', {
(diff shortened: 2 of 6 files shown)


Psheikomaniac:
### Code review

Found 1 issue:

1. The size-based first-load guard (`previousIds.size === 0`) silenced the very first new reservation on a dashboard that started empty: the second watcher fire still saw size 0 and was treated as another first-load. Replaced by an explicit `hasBaseline` flag and added a regression test.

https://github.com/Psheikomaniac/reservation-agent/blob/66494cfe124bf4ae1e9f5b3bc5f0271cc12c11ca/resources/js/composables/useReservationDiffTrigger.ts#L47-L70

Fixed in cb8ab36.

PR #302, opened 2026-04-30
Add SendDailyDigestJob with hourly tick + timezone-aware shouldSendNow (#249)

Psheikomaniac:
## Summary

Implements PRD-010 § Email-Digest job:

- `app/Jobs/SendDailyDigestJob.php` — `final` class implementing `ShouldQueue`.
- `handle()` loads every user with a `restaurant_id`, eager-loads the restaurant, and filters in-memory to those with `notification_settings.daily_digest === true`.
- `shouldSendNow($user)` compares `Carbon::now($user->restaurant->timezone)->format('H:i')` against the configured `daily_digest_at`. Hourly cadence + per-user filter is the simplest robust approach for multi-timezone deployments.
- **Idempotency**: `Cache::add('digest-sent:{user_id}:{YYYY-MM-DD}', true, 23h)`. Atomic check-and-set; a parallel worker that races simply gets `false` and skips. The 23-hour TTL clears before tomorrow's window opens.
- Mail dispatched via `DigestSummary::forUser($user, route('dashboard'))` + `Mail::to($user)->send(new DailyDigestMail(...))`.

## Why

Hourly + per-user clock filter means a single cron line covers every restaurant timezone — no cron-per-zone configuration. The cache lock makes the job safe to invoke twice within a minute (e.g. CI re-tests, replay).

## Test plan

- [x] `php artisan test --filter='DailyDigest'` — 11 passed (mail + job)
- [x] `php artisan test` — 816 passed
- [x] `./vendor/bin/pint --test` — clean

Test cases cover: restaurant-tz match, off-by-30min skip, `daily_digest = false` skip, no `restaurant_id` skip, idempotency (twice in same minute → 1 send + lock present), and two restaurants in different timezones firing in their respective windows.

Closes #249

PR #301, opened 2026-04-30
Add DailyDigestMail + DigestSummary DTO (#250)

Psheikomaniac:
## Summary

Introduces the email half of PRD-010 § Email-Digest:

- `app/Services/Notifications/DigestSummary.php` — `final readonly` DTO with `restaurantName`, `totalToday`, `confirmed`, `pending`, `needsReview`, `dashboardUrl`. Static `forUser($user, $dashboardUrl)` factory aggregates counts scoped by `restaurant_id` so multi-restaurant chains never leak.
- The "today" window uses the **restaurant timezone**, not the server clock — the 18:00 digest in Berlin must cover everything since midnight Berlin even when the scheduler ticks in UTC.
- `app/Mail/DailyDigestMail.php` — Mailable carrying the DTO. Subject `Tageszusammenfassung — {restaurant_name}` per spec.
- `resources/views/emails/notifications/daily-digest.blade.php` — HTML view: restaurant header, four numbered cards (Heute gesamt / Bestätigt / Offen / Manuelle Prüfung), dashboard CTA, footer with a link to `settings.notifications.edit` so the operator can opt out from the mail itself.

## Why

PRD-010 demands the digest as the fallback for users that block browser notifications. Aggregating in a DTO keeps the queue worker thin — the calculation runs on the worker; the mailable just renders.

## Test plan

- [x] `php artisan test --filter='DailyDigest'` — 5 passed (subject, counts today, multi-tenant isolation, timezone window, view render with URL + numbers + settings link)
- [x] `php artisan test` — 810 passed
- [x] `./vendor/bin/pint --test` — clean

Closes #250

PR #297, opened 2026-04-30
useNotifications composable for dashboard push/sound (#245)

Psheikomaniac:
## Summary

- New \`resources/js/composables/useNotifications.ts\` wraps the Browser Notifications API and \`HTMLAudioElement\` for the dashboard's PRD-010 layer. Returns \`{ permission, requestPermission, notify, playSound }\`.
- Defensive on three axes:
  1. Missing API (test environments / older Safari): \`Notification\` / \`Audio\` may be undefined. The composable falls back to \`'denied'\` and no-ops instead of throwing.
  2. User opt-out: \`notify\` is silent when \`settings.browser_notifications === false\`; \`playSound\` is silent when \`settings.sound_alerts === false\`. The dashboard polling loop can call them unconditionally.
  3. Browser autoplay policy: \`audio.play()\` rejection is intentionally swallowed (no UI error).
- Volume clamped to \`[0, 1]\` so a corrupt \`volume\` JSON value can't blow out the user's speakers.
- New \`NotificationSettings\` interface in \`resources/js/types/index.ts\`.

## Why

Issue #245 — the frontend wrapper that #248 (dashboard polling diff trigger) and #246 (settings page test-sound button) will both call.

Closes #245

## Test plan

- [x] \`npx vitest run resources/js/composables/useNotifications.test.ts\` (9 passed)
- [x] \`npm run test\` (65 passed; the existing composable suites still green)
- [x] \`npm run lint\` + \`npm run format:check\`
- [x] \`php artisan test\` (793 passed; backend unaffected)
- [x] \`./vendor/bin/pint --test\`

Test cases: opt-out short-circuits; permission-not-granted short-circuit; happy-path notification; missing Notification API returns \`denied\`; \`requestPermission\` updates the ref; \`audio.play\` rejection is swallowed; sound_alerts opt-out skips Audio instantiation; volume clamp on \`>100\` / \`<0\` / mid range; soundKey override picks the right \`/sounds/{key}.mp3\` path.

PR #304, opened 2026-04-30
Add cross-cutting digest-only fallback test (#252)

### LINKED ISSUE #252: Tests: composable, dashboard diff, daily digest, settings (consolidation)

Psheikomaniac:
## Summary

Issue #252 is the consolidation gate that verifies every PRD-010 acceptance criterion has test coverage. All other listed cases were already covered by sibling PRs:

| Case | File | PR |
|------|------|----|
| `useNotifications.test.ts` (composable behaviours) | `resources/js/composables/useNotifications.test.ts` | #297 |
| Diff-trigger initial load / new ids / suppressed when off / filter change | `resources/js/composables/useReservationDiffTrigger.test.ts` | #300 |
| Digest at user time in restaurant timezone | `tests/Feature/Notifications/DailyDigestJobTest.php` | #302 |
| Digest off → no send | same | #302 |
| Once per user per day | same | #302 |
| Skips users without `restaurant_id` | same | #302 |
| Counts for today | `tests/Feature/Notifications/DailyDigestMailTest.php` | #301 |
| Includes dashboard URL | same | #301 |
| Multi-tenant isolation | same | #301 |
| `SettingsTest.php` (persists / merges defaults / forbids cross-user) | `tests/Feature/Notifications/SettingsTest.php` | #298 |

The single remaining cross-cutting case from the consolidation list — `falls back to digest only when Notification API is undefined` — is added in this PR. It deletes both `window.Notification` and `window.Audio`, asserts the composable reports `denied`, and proves both `notify()` and `playSound()` no-op silently so the dashboard polling cannot crash on a missing global.

## Why

Without this guard, an older browser would surface a `ReferenceError` on the first poll cycle and tear down the entire dashboard refresh — losing not only notifications but also the row updates the digest is meant to back up.

## Test plan

- [x] `npx vitest run` — 82 tests passing
- [x] `php artisan test` — 817 passing
- [x] `npm run lint`, `npm run format:check`, `./vendor/bin/pint --test` — clean

Closes #252

### DIFF
--- a/resources/js/composables/useNotifications.test.ts
+++ b/resources/js/composables/useNotifications.test.ts
@@ -183,4 +183,49 @@ describe('useNotifications', () => {
 
         expect(FakeAudio.lastInstance?.src).toBe('/sounds/chime.mp3');
     });
+
+    /**
+     * PRD-010 § Risiken & offene Fragen — "Browser-Block": when the browser
+     * has no Notification API at all (older Safari, embedded webviews, some
+     * test runners), the composable must stay completely silent — every
+     * call has to no-op without throwing — so the digest email becomes the
+     * only delivery channel for the operator. This is the consolidation
+     * gate from issue #252.
+     */
+    it('falls back to digest only when Notification API is undefined', () => {
+        // Wipe both sides of the browser API: notifications and audio
+        // can be missing independently in real environments, but for the
+        // "fallback to digest" path we model the worst case where the
+        // tab has neither.
+        Reflect.deleteProperty(window, 'Notification');
+        Reflect.deleteProperty(window, 'Audio');
+
+        const settings = ref(
+            makeSettings({
+                browser_notifications: true,
+                sound_alerts: true,
+                daily_digest: true, // digest stays the only channel
+            }),
+        );
+
+        const handle = useNotifications(settings);
+
+        // Permission is reported as denied — the settings page uses this
+        // signal to disable the toggle and show the "blocked" hint.
+        expect(handle.permission.value).toBe('denied');
+
+        // Both side-effect functions must no-op silently. If either threw
+        // a "ReferenceError: Notification is not defined" the dashboard
+        // diff trigger would crash on the first poll cycle and the user
+        // would lose every notification path *including* the polling
+        // refresh that drives the row updates.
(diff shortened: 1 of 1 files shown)


Psheikomaniac:
### Code review

Found 2 issues:

1. The new test redundantly re-asserted `permission.value === 'denied'`, already covered by an earlier case in the same file (`'returns denied permission when Notification API is undefined'`).

2. The inline comment on the `FakeAudio.lastInstance` assertion implied that deleting `window.Audio` is what nulls the static field, when it is `beforeEach` that resets the class-level state.

https://github.com/Psheikomaniac/reservation-agent/blob/ccae9b2828bcb4dd9ee20b59b1de8ea9bf9bc6ce/resources/js/composables/useNotifications.test.ts#L188-L230

Both resolved in a490079: dropped the duplicate permission assertion + the now-dangling `lastInstance` check, rewrote the docblock to call out the relationship to the prior test instead of misrepresenting the static-state invariant.

PR #328, opened 2026-05-22
PRD-011: SlotAvailability::freeTablesAt + alternative slot suggestions (#315)

Psheikomaniac:
## Summary

Completes the availability service surface for PRD-011 (plan Task 7):

- `freeTablesAt(restaurantId, datetime): Collection<int, Table>` — the active, unbooked tables for a buffered slot, ordered by `sort_order`/`id` for deterministic output.
- `forSlot` now populates `alternativeSlots` with up to three later **same-day** slots (30-minute steps within opening hours) when the requested slot is `full`. Closed/full candidates are skipped naturally because `evaluateSlot` reports them as full.
- Refactor: `forSlot` is split into the public entry point and a private `evaluateSlot`, so the alternative search reuses slot evaluation without recursion.
- 4 new tests (free tables exclude busy + inactive; alternatives `[22:00, 22:30]` after a full 19:00; cap at three `[17:00, 17:30, 18:00]`; free slot keeps alternatives empty).

## Why

When a caller's desired time is full, the consumer surfaces (the Belegungs-Tab in #321 and PRD-012's phone/walk-in entry) need ready next-best times to offer the guest — still computed deterministically by Laravel, never by the AI.

## Notes

- Inherits the determinism (`withoutGlobalScopes`, ordered tables) and half-open occupancy window from #313/#314 via the shared `activeTables()`/`busyTableIds()` helpers — `freeTablesAt` uses `activeTables()` rather than the raw relationship (the plan sketch omitted `withoutGlobalScopes`, which would have re-introduced the tenant-scope bug).
- The alternative search is bounded: it stops at three results or end-of-day, and `evaluateSlot` never recurses into it.

## Test plan

- [x] `php artisan test tests/Unit/Availability/SlotAvailabilityTest.php` — 21 pass
- [x] `php artisan test` — 859 pass, no regressions
- [x] `./vendor/bin/pint --test` on touched paths — clean
- [x] No JS/TS changes → ESLint/Prettier not applicable

Closes #315

Issue #315, opened 2026-05-01
PRD-011: SlotAvailability::freeTablesAt + alternative slot suggestions


Psheikomaniac:
## Goal

`freeTablesAt(Restaurant, slot)` returns table ids free at that moment. When `forSlot` returns Full, it populates `alternative_slots` with up to 3 next free slots (same day, 30-min steps, within opening hours). Used by PRD-012's UI.

Refs PRD-011 — `docs/PRD-011-availability-and-tables.md`.
Plan: `docs/superpowers/plans/2026-05-01-prd-011-availability-and-tables.md` § Task 7.

## Files

- Modify: `app/Services/Availability/SlotAvailability.php`
- Modify: `tests/Unit/Availability/SlotAvailabilityTest.php`

## Acceptance criteria

- [ ] `freeTablesAt(Restaurant, Carbon $slot): array` returns table ids free in that slot (buffer-aware, active only)
- [ ] When `forSlot` returns `full`, `alternative_slots` is populated with up to 3 future free slots that day
- [ ] Search walks forward in 30-min increments; stops at end of opening hours
- [ ] If `forSlot` is not full, `alternative_slots` is empty
- [ ] `composer test` green; `pint --test` clean

## Tests

- [ ] freeTablesAt returns correct ids in each scenario (none, some, all)
- [ ] alternativeSlots returns up to 3 results in chronological order
- [ ] Stops at closing time (no slots past opening hours)
- [ ] Empty when day has no later free slots

## Depends on

#314

## Branch

`feature/315-free-tables-and-alternatives` from `dev`.

Psheikomaniac:
Done via PR #328 (functional implementation) and PR #329 (self-review hardening), both merged into `dev` (commits ab0a3b2, e236667). Closing manually (auto-close doesn't fire on `dev` merges).

Acceptance criteria:
- [x] `freeTablesAt(restaurantId, datetime)` returns the active, unbooked tables for a buffered slot (ordered `sort_order`/`id`)
- [x] `forSlot` populates `alternativeSlots` with up to three later same-day slots (30-min steps within opening hours) when the requested slot is full
- [x] `php artisan test` green (860), `pint --test` clean

Notes:
- Self-review surfaced a per-candidate N+1 (restaurant/tables refetched per alternative) → refactored to load once and reuse via `evaluateLoadedSlot`; day boundary now taken in the restaurant-local timezone; class docblock refreshed.
[...]

PR #329, opened 2026-05-22
fix(PRD-011): apply self-review hardening lost from #328 squash

Psheikomaniac:
## Summary

PR #328 (#315) was squash-merged at the branch state **before** its self-review fix commit was included, so three review improvements never reached `dev`. This re-applies them (cherry-pick of the lost commit):

- **Per-candidate N+1:** resolve the restaurant + active tables once in `forSlot` and reuse them for every alternative-search candidate via `evaluateLoadedSlot`, instead of re-running `Restaurant::findOrFail` + the tables query per 30-minute step.
- **Day-boundary timezone:** take the alternative-search end-of-day in the restaurant's local timezone (`endOfDay()`), consistent with the timezone-aware opening-hours check; adds a test that alternatives never roll into the next local day.
- **Stale docblock:** the class docblock no longer says alternative slots "extend this service later" — they are implemented.

## Why

The functional acceptance criteria of #315 are already on `dev` from #328; this restores the review hardening that the squash dropped. See the review thread on #328.

## Test plan

- [x] `php artisan test tests/Unit/Availability/SlotAvailabilityTest.php` — 22 pass
- [x] `php artisan test` — 860 pass, no regressions
- [x] `./vendor/bin/pint --test` on touched paths — clean

Refs #315

### DIFF
--- a/app/Services/Availability/SlotAvailability.php
+++ b/app/Services/Availability/SlotAvailability.php
@@ -29,8 +29,8 @@ use Illuminate\Support\Collection;
  * `[desired_at - buffer, desired_at + slot_duration + buffer)`. Slot duration is
  * a fixed 90 min in V3 (per-restaurant configurability is a documented PRD-011
  * follow-up). A party that fits no single table can be seated on a two-table
- * combination via `combinable_with` (max two tables in V3); alternative slots
- * (#315) extend this service later.
+ * combination via `combinable_with` (max two tables in V3). When a requested slot
+ * is full, `forSlot` also offers up to three later same-day slots as alternatives.
  */
 final class SlotAvailability
 {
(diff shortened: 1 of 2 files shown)


Psheikomaniac:
### Code review

This is a clean cherry-pick of the self-review fixes from #328 (N+1 reduction, local-timezone day boundary, docblock) that the squash merge dropped; the diff was already vetted in the #328 thread. Two sub-threshold considerations were checked and dismissed:
- `$after->setTimezone($restaurant->timezone)->endOfDay()` is correct: the alternative search is intentionally bounded to the restaurant-local day containing the slot, not the input's UTC day.
- The kept day-boundary test uses a UTC restaurant by design — a discriminating non-UTC test would require mixing storage timezones in a way production never does (`desired_at` is persisted as app-timezone wall-clock), so the fix is equivalence-preserving for realistic same-frame usage.

Issue #344, opened 2026-05-22
PRD-012: Reservations/Quick.vue keyboard-first form + live availability


Psheikomaniac:
## Goal

Single-screen, keyboard-first entry page. Live availability via debounced (250 ms) Inertia partial reload (`only: ['availability']`); status banner for free/tight/full; next-3-slots click-to-fill on full; table override dropdown.

Refs PRD-012 — `docs/PRD-012-manual-reservation-entry.md` § Inertia-Page Layout, § Live-Verfügbarkeit.

## Files

- Create: `resources/js/pages/Reservations/Quick.vue`

## Acceptance criteria

- [ ] Smart defaults rendered (today, now+1h on 30-min grid, 2 guests)
- [ ] Tab order: date → time → party → name → phone → mail → note
- [ ] `Ctrl+Enter` submits; `Esc` closes (confirm when form is dirty)
- [ ] Debounced 250 ms availability reload on date/time/party change via `router.reload({ only: ['availability'], preserveState, preserveScroll })` — **no fetch/axios**
- [ ] Banner: ✅ free + suggested table / ⚠️ tight / ❌ full + next 3 free slots
- [ ] Clicking a suggested slot fills date/time
- [ ] Table override dropdown bound to `table_id`
- [ ] `npm run lint`, `npm run format:check` clean

## Tests

- [ ] Covered by Vitest in #346

## Depends on

#342

## Part of

#337

## Branch

`feature/344-quick-vue-page` from `dev`.

Psheikomaniac:
Done — merged to `dev` via #368 (squash). CI green.

**Acceptance criteria:**
- [x] `Reservations/Quick.vue` single-screen form with smart defaults (from controller `defaults` prop)
- [x] Live availability via 250 ms-debounced `router.reload({ only: ['availability'] })` (no fetch/axios)
- [x] Status banner free/tight/full with suggested table / combination; full slot lists clickable alternative times that refill date/time
- [x] Ctrl/Cmd+Enter submits, Esc cancels (confirm when dirty)
- [x] Table override + source via native selects (no select primitive in the UI kit)
- [x] `npm run build`/`lint`/`format:check` clean

**Self-review fixes:** party_size coercion + reload guard, types moved to @/types, select/heading styling aligned, emoji removed (a11y), isComposing guard, single cancel button.

Vitest coverage is the next issue, #346. Next up: #345 (dashboard button).

Issue #343, opened 2026-05-22
PRD-012: QuickReservationController@store with locking + auto table assign


Psheikomaniac:
## Goal

Write path: persist a `confirmed` reservation with `source = phone | walk_in`, auto-assign a table via `SlotAvailability::suggestTableCombination` (or honour a manual `table_id`), all inside a transaction with `lockForUpdate` to close the race against parallel web submits. **No AI pipeline, no mail.**

Refs PRD-012 — `docs/PRD-012-manual-reservation-entry.md` § Controller + FormRequest, § Locking-Strategie.

## Files

- Modify: `app/Http/Controllers/QuickReservationController.php` (`store`)
- Create: `app/Http/Requests/QuickReservationStoreRequest.php`
- Modify: `routes/web.php`

## Acceptance criteria

- [ ] Route `POST /reservations/quick` → `reservations.quick.store`
- [ ] Validation: `source` in [phone, walk_in]; `date` after-or-equal today; `time` H:i; `party_size` 1–20; `guest_name` required ≤200; `guest_phone` nullable regex `[+0-9 ()/-]+`; `guest_email` nullable email; `note` nullable ≤500; `table_id` nullable exists+active for restaurant
- [ ] Stores `status = confirmed`, `created_by_user_id = auth id`
- [ ] Auto-assigns smallest fitting table when no `table_id` given; honours manual override
- [ ] `DB::transaction` + `lockForUpdate()` on the restaurant's tables
- [ ] No `reservation_reply` created; no mail dispatched
- [ ] Validation error (clear message) when no table fits / chosen table busy

## Tests

- [ ] `owner can create phone reservation` / `walk_in reservation`
- [ ] `it auto-assigns smallest fitting table` / `accepts manual table override`
- [ ] `it rejects when chosen table is busy` / `when no table fits`
- [ ] `it locks the slot during transaction` (two parallel requests → one fails)
- [ ] `it does not create a reservation_reply` / `does not dispatch any mail job` (`Mail::fake` + `assertNothingSent`)
- [ ] `user from other restaurant cannot store`

## Depends on

#342

## Part of

#337

## Branch

`feature/343-quick-store-controller` from `dev`.

Psheikomaniac:
Done — merged to `dev` via #367 (squash). CI green.

**Acceptance criteria:**
- [x] `POST /reservations/quick` → `reservations.quick.store` (auth + verified + `can:viewAny,Table`)
[...]

Issue #341, opened 2026-05-22
PRD-012: ReservationSource phone/walk_in + created_by_user_id migration


Psheikomaniac:
## Goal

Schema/enum foundation for manual entry: `ReservationSource` gets `phone` and `walk_in`, and `reservation_requests` gets `created_by_user_id` (who entered it) plus a `note` column if not already present.

Refs PRD-012 — `docs/PRD-012-manual-reservation-entry.md` § Technische Anforderungen.

## Files

- Modify: `app/Enums/ReservationSource.php`
- Create: `database/migrations/*_add_created_by_user_id_to_reservation_requests_table.php`
- (If `note` missing) Create: `database/migrations/*_add_note_to_reservation_requests_table.php`

## Acceptance criteria

- [ ] `ReservationSource` has `Phone = 'phone'` and `WalkIn = 'walk_in'`
- [ ] `reservation_requests.created_by_user_id`: fk users, nullable, `nullOnDelete`
- [ ] `reservation_requests.note`: string/text nullable (only if not already present — verify first)
- [ ] Migrations have working `up()` and `down()`; `migrate:fresh` clean, rollback+forward works
- [ ] Existing suite stays green

## Tests

- [ ] No new test code (schema/enum only). Verified via existing suite + `migrate:fresh`.

## Depends on

PRD-011 (merged).

## Part of

#337

## Branch

`feature/341-source-enum-schema` from `dev`.

Psheikomaniac:
Done — merged to `dev` via #365 (squash `63086ae`).

**Acceptance criteria:**
- [x] `ReservationSource` has `Phone = 'phone'` and `WalkIn = 'walk_in'`
- [x] `reservation_requests.created_by_user_id`: fk users, nullable, `nullOnDelete`
- [x] `reservation_requests.note`: text nullable (was missing — only `message` existed; `note` is the internal owner annotation, distinct from the inbound guest message)
- [x] Migrations up/down; `migrate` → rollback → `migrate` clean; full suite green (891)
- [x] Both columns mass-assignable

**Beyond the original scope (from self-review):** added `created_by_user_id` integer cast, the `createdBy()` BelongsTo relation, and tests (enum round-trip, nullable note, nullOnDelete) — the issue's "no test code" wording was wrong per project convention (CLAUDE.md + #99/#100 reviews).

**Follow-ups for later issues (not blockers here):**
[...]

PR #335, opened 2026-05-22
PRD-011: Belegungs-Tab + day-availability endpoint (#321)

Psheikomaniac:
## Summary

Final PRD-011 piece (Task 13): the Belegung (occupancy) tab and its backing endpoint.

**Backend**
- `TableAvailabilityController` — GET endpoint rendering a deterministic day grid from `SlotAvailability::forDay`. Returns active tables + an `availability` payload (date, per-slot time/state/suggested_table_id, total_capacity, reserved_seats) with `activeTab='availability'`.
- `TableAvailabilityRequest` — `date` (Y-m-d, today-or-future), defaults to today when absent so the tab opens without a param.
- Route `GET /tables/availability` behind `auth` + `verified` + `can:viewAny,Table`, registered before the `{table}` routes.

**Frontend**
- `TableAvailabilityGrid.vue` — tables × 30-min slot rows coloured by `SlotState` (free/tight/full), an occupancy headline, suggested-table label resolution, and a date picker that partial-reloads only the `availability` prop.
- `Tables.vue` — tabs become server-driven navigations (Stammdaten → `tables.index`, Belegung → `tables.availability`); the page renders the grid when the server marks `activeTab='availability'`.
- `SlotState` / `AvailabilitySlot` / `DayAvailability` types added.

### Acceptance criteria
- [x] `TableAvailabilityRequest`: `date` Y-m-d, today-or-future
- [x] Controller returns the `DayAvailability` shape via Inertia, refreshed with `router.reload({ only: ['availability'] })`
- [x] Tenant scope via policy; unauthorized (no restaurant) → 403; data scoped by `RestaurantScope`
- [x] Grid: rows = slots, coloured by `SlotState`, with suggested table
- [x] Date picker switches the day via Inertia partial reload
- [x] `ziggy:generate` run (gitignored output refreshed)
- [x] `composer test`, `npm run lint`, `format:check`, `build`, `test` all green

## Why

PRD-011 needs the occupancy view so operators can see, per 30-min slot, whether the restaurant is free/tight/full for a chosen day. The decision stays deterministic in `SlotAvailability` (#316); this endpoint is a thin Inertia adapter, and the grid only renders the result — consistent with the "Laravel decides, UI displays" principle.

[...]

PR #366, opened 2026-05-22
PRD-012 #342: quick-reservation create endpoint + live availability

### LINKED ISSUE #342: PRD-012: QuickReservationController@create + request + route + props
## Goal

Read path of the quick-entry form: render `Reservations/Quick` with a live `availability` prop (from `SlotAvailability::forSlot`, PRD-011) computed from smart defaults or from the query params, plus the restaurant's active `tables`.

Refs PRD-012 — `docs/PRD-012-manual-reservation-entry.md` § Controller + FormRequest, § Routing.

## Files

- Create: `app/Http/Controllers/QuickReservationController.php` (`create`)
- Create: `app/Http/Requests/QuickReservationCreateRequest.php`
- Modify: `routes/web.php` (auth+verified group)

## Acceptance criteria

- [ ] Route `GET /reservations/quick` → `reservations.quick.create` (auth + verified)
- [ ] Smart defaults when no params: today, `now()->ceilMinutes(30)->addHour()`, party_size 2
- [ ] `availability` prop = `SlotAvailability::forSlot(restaurant_id, datetime, party_size)`
- [ ] `tables` prop = active tables of the user's restaurant, ordered by `sort_order`
- [ ] Tenant scoping via the authenticated user's `restaurant_id` (no cross-tenant leakage)
- [ ] `php artisan ziggy:generate --types-only resources/js/ziggy.d.ts` run for the new route
- [ ] `pint --test` clean

## Tests

- [ ] `it shows quick form with smart defaults`
- [ ] `availability prop reflects query params`
- [ ] `user from other restaurant only sees own tables`

## Depends on

#341

## Part of

#337

## Branch

`feature/342-quick-create-controller` from `dev`.

Psheikomaniac:
## Summary

PRD-012 read path (`#342`), building on the schema from #341:

- `GET /reservations/quick` → `reservations.quick.create` (auth + verified)
- `QuickReservationController@create` + `QuickReservationCreateRequest` (nullable date/time/party_size; party_size capped at `ReservationRequest::MAX_PARTY_SIZE`)
- Smart defaults: today, `now()->ceilMinutes(30)->addHour()`, 2 guests
- `availability` prop from `SlotAvailability::forSlot` (state + suggested table/combination + up-to-3 alternative slots), recomputed from query params for the live preview
- `tables` prop = active tables of the own restaurant (`TableResource`), for the override dropdown
- Ran `php artisan ziggy:generate --types-only resources/js/ziggy.d.ts` (committed)

## Why

The quick form's value is live availability while the operator is on the phone. This issue serves and recomputes that verdict; the Vue page (#344) and the store path (#343) follow.

## Notes / decisions

- **Timezone:** the operator types local time; `forSlot` receives the UTC instant it expects (matching how `desired_at` is stored). Defaults and alternative slots are presented in `restaurant->timezone`. Mirrors `TableAvailabilityController`.
- **No-restaurant guard:** the route has no model binding to scope on, so the controller `abort_if`s when the user has no restaurant (a null restaurant would otherwise 500 on `->timezone`).
- **Past-date validation** is deliberately left to the store path (#343); the read path stays lenient so the live preview never errors on a tz-edge "today".

## Test plan

- [x] `php artisan test` — 897 passed (6 new in `QuickReservationCreateTest`)
- [x] `./vendor/bin/pint --test`, `npm run lint`, `npm run format:check` clean
- [x] ziggy.d.ts regenerated and committed (ziggy-drift job)
- [x] Feature tests cover: guest redirect, smart defaults, query-param reflection, party-size cap, own-restaurant-only tables, no-restaurant forbidden

Closes #342

### DIFF
--- a/app/Http/Controllers/QuickReservationController.php
+++ b/app/Http/Controllers/QuickReservationController.php
@@ -1,0 +1,105 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Http\Controllers;
+
+use App\Http\Requests\QuickReservationCreateRequest;
+use App\Http\Resources\TableResource;
+use App\Models\Restaurant;
+use App\Models\Table;
+use App\Services\Availability\DTOs\SlotAvailabilityResult;
+use App\Services\Availability\SlotAvailability;
+use Carbon\CarbonImmutable;
+use Inertia\Inertia;
+use Inertia\Response;
+
+/**
+ * Quick phone/walk-in reservation entry (PRD-012). This controller serves the
+ * read path: it renders the keyboard-first form and, on every date/time/party
+ * change, recomputes a deterministic availability verdict via
+ * {@see SlotAvailability::forSlot}. The store path lives in #343.
+ *
+ * The operator works in the restaurant's local time, so smart defaults and the
+ * `defaults` prop are local, while `forSlot` receives the UTC instant it expects
+ * (matching how reservations store `desired_at`). The controller stays a thin
+ * Inertia adapter — availability is the service's job and tenant scope comes
+ * from the Table model's global RestaurantScope and the user's own restaurant.
+ */
+final class QuickReservationController extends Controller
+{
+    private const int DEFAULT_PARTY_SIZE = 2;
+
+    public function __construct(
+        private readonly SlotAvailability $availability,
+    ) {}
+
+    public function create(QuickReservationCreateRequest $request): Response
+    {
+        $user = $request->user();
+        // The route is only auth/verified-gated (no model binding to scope on),
+        // so guard the one precondition the rest relies on: a user without a
+        // restaurant has no slots to show and must not reach the form.
+        abort_if($user->restaurant === null, 403);
+        $restaurant = $user->restaurant;
+        $validated = $request->validated();
+
+        // The operator types local time; default to "now, rounded up to the next
+        // 30 minutes, plus an hour" so the form opens on a plausible near slot.
+        $local = isset($validated['date'], $validated['time'])
+            ? CarbonImmutable::parse("{$validated['date']} {$validated['time']}", $restaurant->timezone)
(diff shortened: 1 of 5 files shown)


Psheikomaniac:
### Code review (self-review)

5 review lenses run. 4 in-scope issues addressed in 9526c22; false positives noted.

**Fixed:**

1. Authorization was done in the controller (`abort_if($user->restaurant === null, 403)`) — CLAUDE.md: "Autorisierung via Policy … **niemals** … im Controller." Replaced with the route's `can:viewAny,Table` gate (TablePolicy, which already rejects restaurant-less users), mirroring the identical `tables.availability` endpoint.

https://github.com/Psheikomaniac/reservation-agent/blob/9526c229e6130bd8c03f97d88b5c26d95d4fcc77/routes/web.php#L48-L52

2. Static `reservations/quick` route was registered after the `reservations/{reservation}` wildcard — safe today only because of `whereNumber`. Moved it ahead, matching the static-before-wildcard convention used for `tables/availability`.

3. `isset($validated['date'], $validated['time'])` silently dropped a partially-supplied query (e.g. only `date`) and fell back to the full smart default. Now date and time default independently.

https://github.com/Psheikomaniac/reservation-agent/blob/9526c229e6130bd8c03f97d88b5c26d95d4fcc77/app/Http/Controllers/QuickReservationController.php#L48-L57

4. Added a staff-role access test (recurring gap from #335's `test_staff_may_view_availability`). Suite: 898 passed.

**False positives (not changed):**

- `final` on the FormRequest / test class — the sibling `TableAvailabilityRequest` and `TableAvailabilityTest` are not `final` either; FormRequests/tests follow a non-final convention here.
- `Table::query()` should use `withoutGlobalScopes()->where('restaurant_id', …)` — that pattern is for the service layer (jobs/console, no auth). In an HTTP controller the global `RestaurantScope` is the sanctioned tenant mechanism (CLAUDE.md), and this matches `TableAvailabilityController` exactly.

Issue #350, opened 2026-05-22
PRD-013: WaitlistBanner.vue + filter chip + drawer transition buttons


Psheikomaniac:
## Goal

Frontend for the waitlist: banner component (max 5 shown + "… and N more"), a "Warteliste" filter chip, and status-transition buttons in the detail drawer ("Auf Warteliste"; for waitlisted: "Bestätigen" / "Ablehnen").

Refs PRD-013 — `docs/PRD-013-waitlist-passive.md` § UI, § Status-Übergangs-UI.

## Files

- Create: `resources/js/components/WaitlistBanner.vue`
- Modify: `resources/js/pages/Dashboard.vue` (banner + filter chip)
- Modify: reservation detail drawer component (transition buttons)

## Acceptance criteria

- [ ] Banner renders only when entries exist; click opens the reservation detail drawer
- [ ] Banner shows max 5 with truncation summary
- [ ] Filter chip `value="waitlisted" label="Warteliste"`
- [ ] Drawer: "Auf Warteliste" for `new`/`in_review`; "Bestätigen"/"Ablehnen" for `waitlisted`
- [ ] `npm run lint`, `npm run format:check` clean

## Tests

- [ ] Covered by Vitest in #351 + feature test in #349

## Depends on

#349

## Part of

#338

## Branch

`feature/350-waitlist-banner-vue` from `dev`.

Psheikomaniac:
Done — merged to `dev` via #374 (squash). CI green.

- `WaitlistBanner.vue` (max 5 + "… N weitere", per-entry unique testid, role=status/aria-live), rendered on the dashboard; click opens the reservation drawer.
- "Warteliste" status filter chip + waitlisted badge; `'waitlisted'` added to the TS ReservationStatus union.
- Drawer transition buttons: "Auf Warteliste" (new/in_review) and "Bestätigen"/"Ablehnen" (waitlisted), posting to bulk-status (server-validated transitions).

Self-review: unique testids, aria-live, timezone type. Noted pre-existing: bulk-status `skipped` flash isn't surfaced anywhere (affects the existing toolbar too; buttons are gated to valid transitions). Last PRD-013 issue: #351 (Vitest for WaitlistBanner.vue).

Issue #348, opened 2026-05-22
PRD-013: WaitlistBanner::eligibleNow service + unit tests


Psheikomaniac:
## Goal

Service that returns waitlisted reservations whose desired slot is now free (state ≠ full), capped at 20, ordered by `desired_at` ASC. Uses `SlotAvailability::forSlot` (PRD-011).

Refs PRD-013 — `docs/PRD-013-waitlist-passive.md` § Banner-Logik.

## Files

- Create: `app/Services/Waitlist/WaitlistBanner.php`
- Create: `tests/Unit/Waitlist/WaitlistBannerTest.php`

## Acceptance criteria

- [ ] `eligibleNow(int $restaurantId): Collection`
- [ ] Only `status = waitlisted` and `desired_at >= now()`
- [ ] Excludes slots whose `SlotState` is `Full`
- [ ] Hard cap 20; ordered by `desired_at` ASC
- [ ] Final + readonly-injected `SlotAvailability` (stateless service)
- [ ] Runs in <50 ms for 20 waitlisted entries
- [ ] `pint --test` clean

## Tests

- [ ] `it_lists_only_waitlisted_with_free_slot`
- [ ] `it_excludes_past_desired_at`
- [ ] `it_excludes_full_slots`
- [ ] `it_caps_results_at_20`

## Depends on

#347

## Part of

#338

## Branch

`feature/348-waitlist-banner-service` from `dev`.

Psheikomaniac:
Done — merged to `dev` via #372 (squash). CI green.

- `WaitlistBanner::eligibleNow($restaurantId)`: future-dated waitlisted reservations with a non-full slot, oldest first, capped at 20; stateless, delegates to `SlotAvailability`, `withoutGlobalScopes` + explicit restaurant_id.
- 9 unit tests incl. tight-slot-included, null-desired_at-excluded, multi-slot mix, cap, ordering, tenant scope.
- Self-review: `CarbonImmutable::instance` (no string round-trip), null guard. N+1/<50ms noted as a follow-up (would need a SlotAvailability batch method). Next: #349 (dashboard prop + filter).

Issue #347, opened 2026-05-22
PRD-013: ReservationStatus::Waitlisted + status machine transitions


Psheikomaniac:
## Goal

Add the `waitlisted` status and its allowed transitions. Additive only — no schema change (status is stored as string).

Refs PRD-013 — `docs/PRD-013-waitlist-passive.md` § Datenmodell.

## Files

- Modify: `app/Enums/ReservationStatus.php`
- Modify: status machine (`app/Services/ReservationStatusMachine.php` or model method — follow existing pattern)
- Create/Modify: status-machine unit test

## Acceptance criteria

- [ ] `ReservationStatus::Waitlisted = 'waitlisted'` available + cast on model
- [ ] Allowed: `new → waitlisted`, `in_review → waitlisted`, `waitlisted → confirmed`, `waitlisted → declined`
- [ ] All other transitions to/from `waitlisted` rejected (e.g. `confirmed → waitlisted`)
- [ ] `pint --test` clean

## Tests

- [ ] `it_allows_new_to_waitlisted`
- [ ] `it_allows_waitlisted_to_confirmed` / `_to_declined`
- [ ] `it_rejects_confirmed_to_waitlisted`

## Depends on

PRD-011 (merged).

## Part of

#338

## Branch

`feature/347-waitlisted-status` from `dev`.

Psheikomaniac:
Done — merged to `dev` via #371 (squash). CI green.

- `ReservationStatus::Waitlisted` added; transitions new/in_review → waitlisted and waitlisted → confirmed/declined wired into `allowedNextStates()`; all other to/from-waitlisted transitions forbidden (data-driven matrix + focused test).
- Additive (string column, no migration); enum match stays exhaustive.
- Non-occupying PRD-013 invariant pinned in `SlotAvailabilityTest` + `RestaurantAvailableSeatsAtTest`.

Frontend (TS type, filter chip, badge) deferred to #349/#350 — no path sets waitlisted yet, so no runtime effect. Next: #348 (WaitlistBanner service).

PR #373, opened 2026-05-23
PRD-013 #349: dashboard waitlistBanner prop + waitlisted filter (backend)

### LINKED ISSUE #349: PRD-013: Dashboard waitlistBanner prop + waitlisted filter (backend)
## Goal

Wire the waitlist into the dashboard backend: add `waitlistBanner` to the Inertia props (loaded by the existing 30 s poll) and extend the reservation filter to support the `waitlisted` status.

Refs PRD-013 — `docs/PRD-013-waitlist-passive.md` § Banner-Logik, § Filter im Dashboard.

## Files

- Modify: `app/Http/Controllers/DashboardController.php`
- Modify: reservation filter service (PRD-004)

## Acceptance criteria

- [ ] `waitlistBanner` prop = `WaitlistBanner::eligibleNow(restaurant_id)`, tenant-scoped
- [ ] Prop included in the existing poll-reloadable prop set (no new endpoint/interval)
- [ ] Filter `waitlisted` returns only `status = waitlisted`
- [ ] User from other restaurant gets no banner entries
- [ ] `pint --test` clean

## Tests

- [ ] `banner_appears_after_cancel_frees_slot`
- [ ] `banner_disappears_after_waitlisted_confirmed`
- [ ] `dashboard_filter_shows_waitlisted_only`
- [ ] `user_from_other_restaurant_does_not_see_banner`

## Depends on

#348

## Part of

#338

## Branch

`feature/349-dashboard-waitlist-prop` from `dev`.

Psheikomaniac:
## Summary

PRD-013 #349 (backend), on top of the service from #348:

- `DashboardController` exposes a **`waitlistBanner`** prop = `WaitlistBanner::eligibleNow($restaurantId)`, serialized via `ReservationRequestResource::collection(...)->resolve()` as a plain array (so the page uses `banner.length` directly, like `threadMessages`). Empty when the user has no restaurant.
- The **waitlisted status filter** required no code change: `status.*` already validates with `Rule::enum(ReservationStatus)` and the model's `scopeFilter` does a `whereIn`, so the value added in #347 flows through. Covered by a test so it can't silently regress.

## Why

The dashboard banner (#350) needs the list of waiting guests who could be seated now, and the status filter needs to surface waitlisted entries. This is the server side of both.

## Notes

- The prop is loaded on every dashboard render; the existing 30s poll refreshes it once #350 adds `waitlistBanner` to the reload's `only` set. Method-injected `WaitlistBanner` (stateless service).
- Tenant scope: `eligibleNow` filters by the explicit `restaurant_id`; a cross-tenant test confirms no leakage.

## Test plan

- [x] `php artisan test` — 942 passed (5 new in `WaitlistFlowTest`)
- [x] `./vendor/bin/pint --test` clean
- [x] Covers: banner lists eligible waitlisted, excludes full-slot, disappears after confirm, filter shows only waitlisted, cross-tenant hidden
- [x] No routes/frontend touched → ziggy-drift + JS jobs unaffected

Closes #349

### DIFF
--- a/app/Http/Controllers/DashboardController.php
+++ b/app/Http/Controllers/DashboardController.php
@@ -11,6 +11,7 @@ use App\Http\Resources\ReservationMessageResource;
 use App\Http\Resources\ReservationRequestDetailResource;
 use App\Http\Resources\ReservationRequestResource;
 use App\Models\ReservationRequest;
+use App\Services\Waitlist\WaitlistBanner;
 use App\Support\OpenAiKeyHealth;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Carbon;
@@ -20,10 +21,11 @@ use Inertia\Response;
 
 final class DashboardController extends Controller
 {
-    public function index(DashboardFilterRequest $request): Response
+    public function index(DashboardFilterRequest $request, WaitlistBanner $waitlistBanner): Response
     {
         $validated = $request->validated();
         $selectedId = isset($validated['selected']) ? (int) $validated['selected'] : null;
+        $restaurantId = $request->user()?->restaurant_id;
 
         $filterQuery = Arr::except($request->query(), ['selected']);
         $filters = $filterQuery === []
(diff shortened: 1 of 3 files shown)


Psheikomaniac:
### Code review (self-review)

Two focused lenses (wiring/perf + test coverage). Fixed in 0df9688.

**Fixed:**

1. **Real bug:** `waitlistBanner` was not in the dashboard poll's `only`-set, so it would have been static after load — my PR note was wrong. Added it to `pollOnly()`, so a cancellation that frees a slot surfaces the waiting guest within one poll tick (the PRD-013 AC). One-line frontend change; `#350` then just renders the now-live prop.

https://github.com/Psheikomaniac/reservation-agent/blob/0df9688dc805d220841659f4338e98852b0bca00/resources/js/pages/Dashboard.vue#L453-L461

2. Added the **cancel-frees-slot transition** feature test (the named PRD AC end-to-end: full → cancel → banner appears), the **no-restaurant-user** branch test, and a `filters.status` echo assertion proving the waitlisted value is accepted, not stripped.

**Reviewed, not changed:**

- **N+1 / per-load cost:** `eligibleNow` runs ≤20 `forSlot` calls on each full load and each poll. Same bounded cost noted in #348; a real fix needs a batch method on `SlotAvailability` (its internals), out of scope here. Acceptable for a 30s poll; flagged as a follow-up optimisation.
- Eager prop, `resolve()` plain-array shape, no-restaurant guard, method injection, tenant scope — all confirmed correct.

`php artisan test` 944 passed; `npm run build`/lint/format clean.

Issue #355, opened 2026-05-22
PRD-014: StoreReservationRequest sync-confirm path + lock + fallback


Psheikomaniac:
## Goal

Extend the public store controller (PRD-002) with an additive sync-confirm path: on `proceed`, generate sync reply, create the table assignment (PRD-011), send mail immediately, set `confirmed` + `sync_confirm`, render `ConfirmedSync`. On any failure/timeout, fall through to the unchanged V1 path (`new` + `Submitted`). Holds `lockForUpdate` on the suggested table (race safety vs PRD-012).

Refs PRD-014 — `docs/PRD-014-sync-web-confirm.md` § Controller-Erweiterung.

## Files

- Modify: `app/Http/Controllers/StoreReservationRequestController.php` (or equivalent)

## Acceptance criteria

- [ ] Sync path triggers only when decider returns `proceed`
- [ ] Creates `ReservationTableAssignment` from the suggested table inside the transaction with `lockForUpdate`
- [ ] Sync (non-queued) mail send; on OpenAI timeout or SMTP failure → fall through to V1 path, no UI error
- [ ] Sets `status = confirmed`, reply `sync_confirm = true`, `sent_at`
- [ ] Only `web_form` source uses the sync path (email/phone never)
- [ ] No PII in failure logs

## Tests

- [ ] `sync_confirm_succeeds_and_sends_mail` (`Mail::fake` + `assertSent`)
- [ ] `sync_confirm_falls_back_on_openai_timeout` / `_on_smtp_failure`
- [ ] `sync_confirm_creates_table_assignment`
- [ ] race: two parallel submits for the same table+slot → one sync-confirms, other falls back to `new`
- [ ] `email_source_does_not_use_sync_confirm`

## Depends on

#353, #354

## Part of

#339

## Branch

`feature/355-sync-confirm-store-path` from `dev`.

Psheikomaniac:
Done via #380, merged to `dev` (squash `49fdbde`).

Acceptance criteria:
- [x] Sync path triggers only on decider `proceed` (web_form only; defensive source guard)
- [x] Creates `ReservationTableAssignment` from the in-lock re-resolved combination, inside the transaction with `lockForUpdate` (scoped to the route restaurant)
- [x] Non-queued send; OpenAI timeout or SMTP failure → full rollback → V1 path, no UI error
[...]

PR #379, opened 2026-05-23
PRD-014 #354: OpenAiReplyGenerator::generateSync with 5 s hard timeout

### LINKED ISSUE #354: PRD-014: OpenAiReplyGenerator::generateSync with 5 s hard timeout
## Goal

Synchronous reply generation for the submit path: a `generateSync` method with a 5 s hard timeout on the OpenAI HTTP client, throwing `OpenAiTimeoutException` on timeout (caller falls back to the V1 path). The existing async `generate` (job pipeline) is unchanged.

Refs PRD-014 — `docs/PRD-014-sync-web-confirm.md` § Controller-Erweiterung.

## Files

- Modify: `app/Services/OpenAiReplyGenerator.php`
- Create: `app/Exceptions/OpenAiTimeoutException.php` (if no suitable one exists)

## Acceptance criteria

- [ ] `generateSync(ReservationRequest $r, int $timeout = 5)` sets a hard client timeout
- [ ] Timeout throws `OpenAiTimeoutException`
- [ ] Existing async `generate` behaviour unchanged
- [ ] No API key in returned strings, logs, or exceptions
- [ ] `pint --test` clean

## Tests

- [ ] `it generates a reply synchronously` (`OpenAI::fake`)
- [ ] `it throws OpenAiTimeoutException on timeout`
- [ ] `it never leaks the api key`

## Depends on

PRD-005 (merged).

## Part of

#339

## Branch

`feature/354-openai-generate-sync` from `dev`.

Psheikomaniac:
## Summary

Adds the synchronous reply path for the PRD-014 web sync-confirm flow.

- `OpenAiReplyGenerator::generateSync(array $context, int $timeout = 5): string` — runs a timeout-bounded Chat Completions call.
- New `App\Exceptions\AI\OpenAiTimeoutException`; a transport timeout (openai-php wraps the Guzzle `ConnectException` into `TransporterException`) maps to it.
- **Fail-closed contract** (opposite of async `generate`): a sync confirmation must carry a real AI reply, so `generateSync` never returns the neutral fallback — a timeout throws `OpenAiTimeoutException`, an empty completion throws `RuntimeException`. The caller (#355) falls back to the V1 path.
- The 5 s budget is enforced by an optional `syncClientFactory` closure: bound in `AppServiceProvider` to a Guzzle client built with the sync hard-limit (mirroring the openai-php ServiceProvider construction); left `null` in unit tests so the injected fake is used and `OpenAI::fake()` applies.
- The async `generate` (30 s, fallback-returning, job pipeline) is unchanged; the shared chat-create payload was extracted to remove duplication.
- No API key in returned strings, logs, or exceptions (logging hygiene matches `generate`).

### Signature note
PRD-014 / the issue sketch `generateSync(ReservationRequest $r, …)`, but the established pattern keeps the generator a pure formatter over the deterministic context-array (`generate(array $context)`), DB-free and testable. `generateSync` follows that pattern; the #355 controller builds the context via `ReservationContextBuilder` (exactly as the job does) and passes it in.

## Why

PRD-014 § Controller-Erweiterung: web submits that hit a deterministically free slot should be confirmed within the submit latency. That requires a synchronous, hard-bounded OpenAI call that fails closed instead of returning a placeholder.

## Test plan

- [x] `php artisan test` — 962 passed (3103 assertions); 6 new tests in `OpenAiReplyGeneratorGenerateSyncTest`:
  - generates a reply synchronously
  - throws `OpenAiTimeoutException` on timeout
  - throws (not fallback) on empty completion
  - passes the timeout to the sync client factory + defaults to 5 s
  - API key never leaks into logs or the thrown exception
- [x] `./vendor/bin/pint --test` — pass (project-wide)
- [x] No routes/JS touched → Ziggy/Vitest unaffected

Closes #354

### DIFF
--- a/app/Exceptions/AI/OpenAiTimeoutException.php
+++ b/app/Exceptions/AI/OpenAiTimeoutException.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace App\Exceptions\AI;
 
 use RuntimeException;
+use Throwable;
 
 /**
  * Signals that the synchronous reply generation (PRD-014 web sync-confirm)
@@ -15,11 +16,15 @@ use RuntimeException;
  * owner review) instead of surfacing a UI error. Unlike the async
  * `generate`, the sync path never returns the neutral fallback text — a
  * confirmation must always carry a real AI reply.
+ *
+ * The originating transport exception is chained as `$previous` so the
+ * underlying cURL/Guzzle detail stays available for debugging (it is
+ * never logged with the API key).
  */
 final class OpenAiTimeoutException extends RuntimeException
 {
-    public function __construct(string $message = 'OpenAI sync reply generation timed out.')
+    public function __construct(string $message = 'OpenAI sync reply generation timed out.', ?Throwable $previous = null)
     {
-        parent::__construct($message);
+        parent::__construct($message, 0, $previous);
     }
 }
--- a/app/Exceptions/AI/OpenAiTimeoutException.php
+++ b/app/Exceptions/AI/OpenAiTimeoutException.php
@@ -1,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Exceptions\AI;
+
+use RuntimeException;
+
+/**
+ * Signals that the synchronous reply generation (PRD-014 web sync-confirm)
+ * exceeded its hard timeout budget or otherwise failed transport-side.
+ *
+ * The sync path runs inside the public submit latency, so the caller
+ * catches this and falls back to the V1 path (`status = new`, normal
+ * owner review) instead of surfacing a UI error. Unlike the async
+ * `generate`, the sync path never returns the neutral fallback text — a
+ * confirmation must always carry a real AI reply.
+ */
+final class OpenAiTimeoutException extends RuntimeException
+{
+    public function __construct(string $message = 'OpenAI sync reply generation timed out.')
+    {
+        parent::__construct($message);
+    }
+}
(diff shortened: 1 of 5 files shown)


Psheikomaniac:
### Code review (self-review)

Ran the five-lens review. No finding scored at/above the 80 posting threshold, but several 75-level findings were legitimate and cheap, so I addressed them in 6050038 rather than ship around them:

- **Error-path coverage (CLAUDE.md "test every scenario").** `generateSync` documented 401/429/5xx propagation but had no tests for it. Added classification + tests: 401 → `OpenAiAuthenticationException`, 429 → `OpenAiRateLimitException`, 5xx propagates, plus a key-leak guard on the classified throw path. This also closes the async/sync asymmetry so a 401 on the sync path can drive the admin key-check notification.
- **Exception chaining.** `OpenAiTimeoutException` now chains the originating `TransporterException` as `$previous`.
- **Misleading interface contract.** Scoped the `ReplyGenerator` "never throw" guarantee to `generate` and noted that `generateSync` deliberately throws.
- **Docblock accuracy.** Corrected the class-level logging-hygiene note (the `ErrorException` path also logs the numeric status).

Verified false positive: `buildTimeoutBoundOpenAiClient` "discards `withProject()`/`withBaseUri()` return values" — the openai-php `Factory` methods mutate `$this` and return `$this`; the code mirrors the upstream ServiceProvider exactly.

Deferred (score 50): the "timed out" log label is used for all transport failures within the 5 s budget; acceptable since the caller falls back regardless and timeouts dominate the budget.

Verification: `php artisan test` 966 passed (3109 assertions); `./vendor/bin/pint --test` clean. Head: 6050038469d339926de5af179a6945b8e8b3139f

Issue #357, opened 2026-05-22
PRD-014: PublicForm/ConfirmedSync.vue page


Psheikomaniac:
## Goal

New public confirmation page shown when a sync-confirm succeeds: ✅ confirmed, date/time/party as plain text, masked guest email, and the "automated because the slot was free" footer note. The existing `Submitted.vue` (V1 fallback) is unchanged.

Refs PRD-014 — `docs/PRD-014-sync-web-confirm.md` § Bestätigungsseiten.

## Files

- Create: `resources/js/pages/PublicForm/ConfirmedSync.vue`

## Acceptance criteria

- [ ] Shows confirmed state with date, time, party size
- [ ] Masked email ("m…@example.com")
- [ ] Footer note explaining the automated confirmation
- [ ] No auth layout (public page)
- [ ] `npm run lint`, `npm run format:check` clean

## Tests

- [ ] `confirmed_sync_page_renders_when_proceed` (feature, in #355)

## Depends on

#355

## Part of

#339

## Branch

`feature/357-confirmed-sync-page` from `dev`.

Psheikomaniac:
Done via #382, merged to `dev`.

- [x] Shows confirmed state with date (German `DD.MM.YYYY`), time, party size
- [x] Masked email (server-masked in #355, e.g. `a…@gmail.com`)
- [x] Footer note explaining the automated confirmation
- [x] No auth layout (uses `PublicLayout`, like `Public/Thanks`)
- [x] `npm run lint`, `npm run format:check`, `npm run build` (vue-tsc) clean; Vitest 112 passed

Page path is `Public/ConfirmedSync.vue` (not `PublicForm/…`) to match the shipped public pages. The #355 flow test now verifies the component file exists.

PR #380, opened 2026-05-23
PRD-014 #355: web sync-confirm path in the public store controller

### LINKED ISSUE #355: PRD-014: StoreReservationRequest sync-confirm path + lock + fallback

Psheikomaniac:
## Summary

Extends `PublicReservationController@store` with the PRD-014 synchronous confirmation path (additive — the V1 flow is untouched on every fall-through).

On a `web_form` submit, after creating the request:
1. `WebSyncConfirmDecider::decide` — only `proceed` continues.
2. Generate the reply via `OpenAiReplyGenerator::generateSync` **before** the transaction, so the up-to-5 s OpenAI call never holds the table lock. The generator is resolved lazily so a missing/invalid `OPENAI_API_KEY` degrades to V1 (V1 submits never build the OpenAI client — mirrors `GenerateReservationReplyJob`).
3. In a `DB::transaction`: lock the restaurant's active tables, **re-resolve** the table combination inside the lock (race guard vs PRD-012), create the reply (`sync_confirm = true`), assign each table, send the mail non-queued, record the outbound `ReservationMessage` (threading parity with `SendReservationReplyJob`), flip the request to `confirmed` and the reply to `sent`.
4. Render `Public/ConfirmedSync` (date/time/party + masked email).

Any failure — gate skip, OpenAI timeout, lost-table race, SMTP error — rolls the transaction back and falls through to the V1 path (`new`, `ReservationRequestReceived` dispatched), with a PII-free warning log.

### Naming note
PRD/issue reference `PublicForm/ConfirmedSync` and `Submitted`, but the shipped pages live in `resources/js/pages/Public/` and the V1 flow redirects to `public.reservations.thanks` (`Public/Thanks`). I followed the real structure: success renders `Public/ConfirmedSync` (page in #357), fallback keeps the existing redirect.

## Why

PRD-014 § Controller-Erweiterung: confirm free-slot web reservations within the submit latency to lift online conversion, while every non-ideal case stays on the safe V1 review path.

## Test plan

- [x] `php artisan test` — 974 passed (3167 assertions). New `WebSyncConfirmFlowTest`:
  - succeeds + sends mail; renders `Public/ConfirmedSync`; creates the table assignment
  - falls back on OpenAI timeout; falls back on SMTP failure (transaction rolls back)
  - two submits for the same slot confirm only once (in-lock re-check)
  - V1 path when disabled; email-source request is never sync-confirmed
- [x] `./vendor/bin/pint --test` clean
- [x] No routes/JS touched → Ziggy/Vitest unaffected (ConfirmedSync.vue is #357)

Closes #355

### DIFF
--- a/app/Http/Controllers/PublicReservationController.php
+++ b/app/Http/Controllers/PublicReservationController.php
@@ -4,19 +4,43 @@ declare(strict_types=1);
 
 namespace App\Http\Controllers;
 
+use App\Enums\MessageDirection;
+use App\Enums\ReservationReplyStatus;
 use App\Enums\ReservationSource;
 use App\Enums\ReservationStatus;
 use App\Events\ReservationRequestReceived;
 use App\Http\Requests\StoreReservationRequest;
+use App\Jobs\SendReservationReplyJob;
+use App\Mail\ReservationReplyMail;
+use App\Models\ReservationMessage;
+use App\Models\ReservationReply;
 use App\Models\ReservationRequest;
+use App\Models\ReservationTableAssignment;
 use App\Models\Restaurant;
+use App\Models\Table;
+use App\Services\AI\OpenAiReplyGenerator;
+use App\Services\AI\ReservationContextBuilder;
+use App\Services\AutoSend\WebSyncConfirmDecider;
+use App\Services\Availability\SlotAvailability;
 use App\Support\Timezone;
+use Carbon\CarbonImmutable;
 use Illuminate\Http\RedirectResponse;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Mail;
 use Inertia\Inertia;
 use Inertia\Response;
+use Psr\Log\LoggerInterface;
+use Throwable;
 
 final class PublicReservationController extends Controller
 {
+    public function __construct(
+        private readonly WebSyncConfirmDecider $decider,
+        private readonly ReservationContextBuilder $contextBuilder,
+        private readonly SlotAvailability $availability,
+        private readonly LoggerInterface $logger,
+    ) {}
+
     public function create(Restaurant $restaurant): Response
     {
         return Inertia::render('Public/ReservationForm', [
@@ -29,7 +53,7 @@ final class PublicReservationController extends Controller
         ]);
     }
 
-    public function store(StoreReservationRequest $request, Restaurant $restaurant): RedirectResponse
+    public function store(StoreReservationRequest $request, Restaurant $restaurant): RedirectResponse|Response
     {
         if ($request->filled('website')) {
             return redirect()->route('public.reservations.thanks', $restaurant);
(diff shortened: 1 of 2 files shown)


Psheikomaniac:
### Code review (self-review)

Five-lens review. The reviewers converged on one genuinely high-severity issue, fixed in 50be030:

- **Ghost-confirmation on rollback (high).** Mail was sent inside the transaction but *before* the outbound-message write and the reply/request status updates, and `renderConfirmation` sat inside the try. A post-send DB failure (or a render error) would roll back the DB while the email was already delivered, then the `catch`/fall-through would dispatch a *second* (V1) pipeline on an already-mailed guest. Fix: persist every DB side-effect first and make `Mail::send` the **last** statement in the transaction (SMTP failure still rolls back cleanly, per PRD-014); render **after** the committed transaction, outside the try. Added an assertion that the outbound message is rolled back on SMTP failure. The only residual window is a DB commit failure *after* a successful send — the irreducible DB+mail dual-write, documented in-code.
- **State-machine bypass / "parity" wording (minor).** Documented the `forceFill` new→confirmed bypass (same trick `SendReservationReplyJob` uses) and scoped the "parity with SendReservationReplyJob" comment to the outbound-message bookkeeping only.

Verified false positives:
- Lock query `where('restaurant_id', …)` is **not** the forbidden tenant-scope pattern — it's the route restaurant on a public, unauthenticated endpoint (RestaurantScope no-ops without a user); the CLAUDE.md rule targets `auth()->user()->restaurant_id` scoping.
- `app(OpenAiReplyGenerator::class)` late-resolution is the established graceful-degradation pattern (GenerateReservationReplyJob); double-pipeline-after-success is guarded by the early `return`; the table-combination loop assigns every table id (no double-bookable second table).

Deferred (owner call, non-blocking): CLAUDE.md's Verbote literally says "kein Auto-Send-Pfad ohne `AutoSendDecider`"; this path is governed by the parallel `WebSyncConfirmDecider` (owner opt-in + hard gates + killswitch), which PRD-014 §Risiken explicitly deemed sufficient. A one-line CLAUDE.md clarification naming `WebSyncConfirmDecider` would remove the letter-vs-spirit gap, but I won't edit the governance doc unilaterally.

Verification: `php artisan test` 974 passed (3168 assertions); `./vendor/bin/pint --test` clean. Head: 50be030d15237726c97b930cd5782228839a8c44

